PR #33, opened 2017-03-24
Added xsense-mvn dependency in Appveyor

diegoferigo:


diegoferigo:
Waiting the fix of robotology/yarp#1132

PR #34, opened 2017-03-24
urdf files updated and general config files added

MartaLorenzini:


francesco-romano:
Please, do not commit the `xml` version.

Issue #53, opened 2017-04-05
Documentation generation broken


francesco-romano:
It seems after docker images

PR #54, opened 2017-04-05
Fixed doxygen generation

diegoferigo:
The dependencies' folders were cloned into $TRAVIS_BUILD_DIR, hence doxygen was seeing them as part of this repo tree.

Fixes issue #53


francesco-romano:
Just a curiosity. 
Why instead of removing the sources (which may take some time slowing down the compilation), not moving to the previous folder as we did in all the others (other repositories) travis configuration.
Something like
`cd ${TRAVIS_BUILD_DIR}/../` before cloning `iDynTree` ?

diegoferigo:
Because travis uses the same script for building dependencies either in the `linux` docker container, and `osx`. Inside the docker container only the cloned git folder is mounted, and the sources are cloned inside. I would't mind of the time `rm` takes though.

francesco-romano:
ok

PR #68, opened 2017-06-19
Updated travis for new YARP release

francesco-romano:


francesco-romano:
Merging as I was forgetting about this

Issue #126, opened 2019-06-10
Add Windows CI


lrapetti:
The Windows `AppVeyor` CI has been disabled (see https://github.com/robotology/human-dynamics-estimation/issues/60#issuecomment-500336621). 

So it has to be added back, probably using Trevis in order to have a single CI system.

lrapetti:
It would be good to have it in order to avoid regressions such as https://github.com/robotology/human-dynamics-estimation/issues/171

lrapetti:
We can take care of this while addressing https://github.com/robotology/human-dynamics-estimation/issues/217.

lrapetti:
Closing via https://github.com/robotology/human-dynamics-estimation/pull/221

PR #133, opened 2019-06-26
Fix devel windows compilation

lrapetti:
This PR fixes the compilation of the `devel` branch on Windows:
- add `_USE_MATH_DEFINES` definition (with the solution2 proposed in https://github.com/robotology/QA/issues/81)
- remove variable length arrays (related discussion https://stackoverflow.com/questions/33423502/expression-did-not-evaluate-to-a-constant-c)


lrapetti:
The Travis check are successful (N.B. the checks are only for Ubuntu and MacOS, not for Windows see https://github.com/robotology/human-dynamics-estimation/issues/126).

The Codacy failed since it seems to be not enabled for this branch. @diegoferigo it seems only you can enable it, can you procede? Otherwise, we just remove the Codacy check.

Merging

PR #136, opened 2019-07-30
Feature: Add base transform fixed position and orientation parameter to human state publisher

lrapetti:
In order to enforce the fixed base (or even better the fixed base position), I have add a parameter to the [`HumanStatePublisher`](https://github.com/robotology/human-dynamics-estimation/tree/devel/publishers/HumanStatePublisher) that allows to fix the base position or orientation to a given configuration. In this way it will be possible to adjust its position in order to be coherent with the other model position (given by calibrating xsens) in multiple models scenario.

### DIFF
--- a/publishers/HumanStatePublisher/HumanStatePublisher.cpp
+++ b/publishers/HumanStatePublisher/HumanStatePublisher.cpp
@@ -57,6 +57,8 @@ public:
     hde::interfaces::IHumanState* humanState = nullptr;
 
     bool firstRun = true;
+    bool fixBasePosition = false;
+    bool fixBaseOrientation = false;
     std::string baseTFName;
 
     // Buffers
@@ -88,6 +90,36 @@ bool HumanStatePublisher::open(yarp::os::Searchable& config)
     if (!(config.check("period") && config.find("period").isFloat64())) {
         yInfo() << LogPrefix << "Using default period: " << DefaultPeriod << "s";
         useDefaultPeriod = true;
+    }
+
+    yarp::os::Bottle* fixedBasePosition;
+    if (config.check("fixBasePosition")) {
+        if (config.find("fixBasePosition").isList() &&
+                config.find("fixBasePosition").asList()->size() == 3) {
+            pImpl->fixBasePosition = true;
+            fixedBasePosition = config.find("fixBasePosition").asList();
+            yInfo() << LogPrefix << "Using a fixed position for the base frame: "
+                    << fixedBasePosition;
+        }
+        else {
+            yError() << LogPrefix << "Parameter 'fixBasePosition' invalid";
+            return false;
+        }
+    }
+
+    yarp::os::Bottle* fixedBaseOrientation;
+    if (config.check("fixBaseOrientation")) {
+        if (config.find("fixBaseOrientation").isList() &&
+                config.find("fixBaseOrientation").asList()->size() == 4) {
+            pImpl->fixBaseOrientation = true;
+            fixedBaseOrientation = config.find("fixBaseOrientation").asList();
+            yInfo() << LogPrefix << "Using a fixed orientation for the base frame: "
+                    << fixedBaseOrientation;
+        }
+        else {
+            yError() << LogPrefix << "Parameter 'fixBaseOrientation' invalid (required quaternion)";
+            return false;
+        }
     }
 
     // ROS TOPICS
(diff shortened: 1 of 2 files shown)


publishers/HumanStatePublisher/HumanStatePublisher.cpp
```diff
@@ -90,6 +92,36 @@ bool HumanStatePublisher::open(yarp::os::Searchable& config)
         useDefaultPeriod = true;
     }
 
+    yarp::os::Bottle* fixedBasePosition;
+    if (config.check("fixBasePosition")) {
```
Yeshasvitvs: @lrapetti having the nested if conditions such as here looks cumbersome. Try to clean this section to have minimum number  of conditional logic. Usually it is clean to have a parameter check first and then parsing the parameters at a later stage during the initialization. For reference please check https://github.com/robotology/human-dynamics-estimation/blob/40e949a4e09aad4a412b0f7af107d80d7831ce7b/devices/HumanStateProvider/HumanStateProvider.cpp#L218 and https://github.com/robotology/human-dynamics-estimation/blob/40e949a4e09aad4a412b0f7af107d80d7831ce7b/devices/HumanStateProvider/HumanStateProvider.cpp#L304

publishers/HumanStatePublisher/HumanStatePublisher.cpp
```diff
@@ -250,9 +295,14 @@ void HumanStatePublisher::run()
     // PREPARE THE BASE POSITION MESSAGE
     // =================================
 
-    // Get the data from the interface
-    pImpl->humanStateBuffers.basePosition = pImpl->humanState->getBasePosition();
-    pImpl->humanStateBuffers.baseOrientation = pImpl->humanState->getBaseOrientation();
+    // Get the data from the interface if not using fixed values
+    if (!pImpl->fixBasePosition) {
```
Yeshasvitvs: Instead of a separate `if` conditional check, it is better to use `else` condition in conjunction with the previous check of `fixBasePosition` or `fixBaseOrientation`

publishers/HumanStatePublisher/HumanStatePublisher.cpp
lrapetti: The reason for this structure is that the `fixBasePosition` option is an optional flag.
So, if the configuration is present we check if its value is valid, otherwise, we simply skip the option (this because most of the time this option is not used).

publishers/HumanStatePublisher/HumanStatePublisher.cpp
lrapetti: The two options are not exclusive, it can be that we are fixing the orientation and not the position, or the other way around (e.g. what we did for the video). So the two conditions are independent.

Yeshasvitvs: I mean to use 

```

if (pImpl->fixBaseOrientation) {
        pImpl->humanStateBuffers.baseOrientation[0] = fixedBaseOrientation->get(0).asFloat64();
        pImpl->humanStateBuffers.baseOrientation[1] = fixedBaseOrientation->get(1).asFloat64();
        pImpl->humanStateBuffers.baseOrientation[2] = fixedBaseOrientation->get(2).asFloat64();
        pImpl->humanStateBuffers.baseOrientation[3] = fixedBaseOrientation->get(3).asFloat64();
    }
else {
        pImpl->humanStateBuffers.basePosition = pImpl->humanState->getBasePosition();
} 

```

and similarly for `fixBaseOrientation`

publishers/HumanStatePublisher/HumanStatePublisher.cpp
Yeshasvitvs: Makes sense in the case of optional parameter.

publishers/HumanStatePublisher/HumanStatePublisher.cpp
Yeshasvitvs: @lrapetti Sorry I was saying something stupid. The logic is correct, the if `fixBasePosition` or `fixBaseOrientation` parameters are present, the base position if updated accordingly once in the configuration and if the parameters are absent, the base position is updated on every call to the run method.

Yeshasvitvs:
I will proceed to merge this to the devel branch. 

@lrapetti @kouroshD

Issue #152, opened 2019-10-08
Remove base position offset in human-dynamic-estimation


lrapetti:
The idea is to add rpc functionality to the human-state-provider in order to be able to remove an offset for the base position. In this way it would be possible to take care of the relative transform between two humans.

Given some deadline requirements, this feature will be first implemented on top of the following feature branch: [`feature/visualize-berdy-estimated-wrench`](https://github.com/robotology/human-dynamics-estimation/tree/feature/visualize-berdy-estimated-wrench)

lrapetti:
CC @prashanthr05 @Yeshasvitvs @kouroshD

lrapetti:
It has been implemented in https://github.com/lrapetti/human-dynamics-estimation/commit/08b75c27b718e2ab632b482146188a26ff36ec28.

It allows to remove the position offset for one or more axis, and the rotation offset for one or more axis. Preliminary test have been completed, further testing will follow,

lrapetti:
PR opened to the feature branch (https://github.com/robotology/human-dynamics-estimation/pull/153)

claudia-lat:
In the meanwhile I'm moving this issue onto the Iteration.

lrapetti:
PR has been merged, closing.

PR #153, opened 2019-10-11
Add base pose offset

lrapetti:
This PR add the feature for removing position offsets (https://github.com/robotology/human-dynamics-estimation/issues/152) + configuration files for running two human state provider.

The offset is removed by sending command to an `rpc` port opened by the `HumanStateProvider` device. It is possible to remove both position and orientation offsets for one or multiple axis.


lrapetti:
@Yeshasvitvs if you agree I would merge this new feature to this WIP branch so that then it will be merged together in the `master`.

PR #166, opened 2019-12-06
Fix wrench estimation directed as the gravity

lrapetti:
With this PR we fix the estimated wrenches at the hands to be directed as the gravity.


lrapetti:
we can merge in this feature brench and clean the commit later.

PR #181, opened 2020-02-17
Add matio logging to HumanDataCollector device

Yeshasvitvs:
#### Major Features

This PR mainly adds the feature of logging human data from `HumanDataCollector` device https://github.com/robotology/human-dynamics-estimation/pull/178 to a `.mat` file using [matio library](https://github.com/tbeu/matio) as discussed in https://github.com/robotology/human-dynamics-estimation/issues/179. 

#### Minor Fixes
- Update `YARP_OS` to `YARP_os` in cmake files to remove cmake warning https://github.com/robotology/human-dynamics-estimation/pull/181/commits/33c6ada4314ee078f58125593b058207e0a18369

### DIFF
--- a/devices/HumanDataCollector/HumanDataCollector.cpp
+++ b/devices/HumanDataCollector/HumanDataCollector.cpp
@@ -25,6 +25,7 @@
 #include <chrono>
 #include <ratio>
 #include <algorithm>
+#include <utility>
 #include <unordered_map>
 
 const std::string DeviceName = "HumanDataCollector";
@@ -33,7 +34,7 @@ constexpr double DefaultPeriod = 0.01;
 
 using namespace hde::devices;
 
-void writeVectorOfStringToMat(const std::string name, const std::vector<std::string>& strings, mat_t* mat)
+void writeVectorOfStringToMatCell(const std::string name, const std::vector<std::string>& strings, mat_t* matFile)
 {
     if (strings.empty()) {
         yError() << LogPrefix << "Passed string is empty";
@@ -41,12 +42,12 @@ void writeVectorOfStringToMat(const std::string name, const std::vector<std::str
     }
 
     size_t dims[2] = {strings.size(), 1};
-    matvar_t* cell_array = Mat_VarCreate(name.c_str(), MAT_C_CELL, MAT_T_CELL, 2, dims, nullptr, 0);
+    matvar_t* matCellArray = Mat_VarCreate(name.c_str(), MAT_C_CELL, MAT_T_CELL, 2, dims, nullptr, 0);
 
-    if (cell_array == nullptr) {
+    if (matCellArray == nullptr) {
         yError() << LogPrefix << "Failed to create mat cell array " << name;
-        Mat_VarFree(cell_array);
-        Mat_Close(mat);
+        Mat_VarFree(matCellArray);
+        Mat_Close(matFile);
         return;
     }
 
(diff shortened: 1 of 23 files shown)


devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -1211,7 +1198,7 @@ bool HumanStateProvider::open(yarp::os::Searchable& config)
         }
     }
     else {
-        yInfo() << "CUSTOM CONSTRAINTS are not defined in xml file.";
+        //yInfo() << "CUSTOM CONSTRAINTS are not defined in xml file.";
```
lrapetti: this `else` condition is now empty, can we remove it?

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -1235,7 +1222,7 @@ bool HumanStateProvider::open(yarp::os::Searchable& config)
                  << pImpl->customConstraintMatrix.cols() << ") are not equal";
         return false;
     }
-    yInfo() << "******* DOF: " << modelLoader.model().getNrOfDOFs();
+    //yInfo() << "******* DOF: " << modelLoader.model().getNrOfDOFs();
```
lrapetti: this commented line can be removed

devices/HumanDataCollector/HumanDataCollector.cpp
```diff
@@ -98,6 +238,18 @@ class HumanDataCollector::impl
 
     yarp::os::BufferedPort<yarp::os::Bottle> dynamicsJointNamesDataPort;
     yarp::os::BufferedPort<yarp::sig::Vector> jointTorquesDataPort;
+
+
+    // MATIO Logging
+    bool matioLogger = false;
+    const std::experimental::filesystem::path originalWorkingDirectory = std::experimental::filesystem::current_path();
+    std::string matLogDirectory = "";
+    std::string matLogFileName = "matLogFile.mat"; //TODO: Improve file handling for multiple file logging
```
lrapetti: the name of the file might be also passed as a configuration

devices/HumanDataCollector/HumanDataCollector.cpp
```diff
@@ -286,13 +456,79 @@ void HumanDataCollector::run()
...
 
+        // Initialize  matio logging
+        if (pImpl->matioLogger) {
+
+            // Get the start time
+            pImpl->startTime = std::chrono::system_clock::now();
+
+            // TODO: Handle multiple files
+            std::string matLogFileFullPathName = pImpl->matLogDirectory + pImpl->matLogFileName;
+            pImpl->matFilePtr = Mat_CreateVer(matLogFileFullPathName.c_str(), nullptr, MAT_FT_MAT73);
```
lrapetti: what happen here if the file already exists? is it overwritten or is it crashing (I see the multiple file handling is still a TODO, I was just curios about the current behaviour)

Yeshasvitvs: In the current implementation, if the file already exists, it  will be rewritten without causing any errors. One way to proceed for multiple file handling is to append the files with data and time.

devices/HumanDataCollector/HumanDataCollector.cpp
```diff
@@ -98,6 +237,18 @@ class HumanDataCollector::impl
 
     yarp::os::BufferedPort<yarp::os::Bottle> dynamicsJointNamesDataPort;
     yarp::os::BufferedPort<yarp::sig::Vector> jointTorquesDataPort;
+
+
+    // MATIO Logging
+    bool matioLogger = false;
+    const std::experimental::filesystem::path originalWorkingDirectory = std::experimental::filesystem::current_path();
```
diegoferigo: Maybe this is one of those cases where `auto` does not affect readability :)

devices/HumanDataCollector/HumanDataCollector.cpp
```diff
@@ -331,12 +566,93 @@ void HumanDataCollector::run()
...
+        }
+
+        if (pImpl->isAttached.wrenchProvider) {
+
+            // Set wrench measurements source names once
+            if (pImpl->humanDataStruct.wrenchMeasurementSourceNames.empty()) {
+                pImpl->humanDataStruct.wrenchMeasurementSourceNames = pImpl->wrenchMeasurementSourceNames;
+            }
+
+            pImpl->humanDataStruct.data["wrenchMeasurements"].push_back(pImpl->wrenchMeasurementValuesVec);
```
diegoferigo: Sometimes you use `at()` and other you use the `[]` operator. Is there any reason?

devices/HumanDataCollector/HumanDataCollector.cpp
```diff
@@ -551,6 +851,45 @@ bool HumanDataCollector::detach()
...
+    if (pImpl->matioLogger) {
+
+        // Set the string variables elements as cell arrays to mat cell variables
+        writeVectorOfStringToMatCell("stateJointNames", pImpl->humanDataStruct.stateJointNames, pImpl->matFilePtr);
+        writeVectorOfStringToMatCell("wrenchMeasurementSourceNames", pImpl->humanDataStruct.wrenchMeasurementSourceNames, pImpl->matFilePtr);
+        writeVectorOfStringToMatCell("wrenchEstimateSourceNames", pImpl->humanDataStruct.wrenchEstimateSourceNames, pImpl->matFilePtr);
+        writeVectorOfStringToMatCell("dynamicsJointNames", pImpl->humanDataStruct.dynamicsJointNames, pImpl->matFilePtr);
+
+        // Correct the time values stored in the time vector to zero start value
+        std::transform( pImpl->humanDataStruct.time.begin(), pImpl->humanDataStruct.time.end(), pImpl->humanDataStruct.time.begin(), std::bind2nd( std::plus<long>(), - pImpl->humanDataStruct.time.at(0) ) );
```
diegoferigo: How could this super long line trick clang format?

devices/HumanDataCollector/HumanDataCollector.cpp
```diff
@@ -16,18 +16,137 @@
...
+#include <utility>
+#include <unordered_map>
+
 const std::string DeviceName = "HumanDataCollector";
 const std::string LogPrefix = DeviceName + " :";
 constexpr double DefaultPeriod = 0.01;
 
 using namespace hde::devices;
 
+void writeVectorOfStringToMatCell(const std::string name, const std::vector<std::string>& strings, mat_t* matFile)
```
diegoferigo: I'm not fully conviced that you handle the errros properly. I see a lot of `Mat_Close` and no error propagation from these `write*` functions to the outside. You call this method sequentally multiple times around line 850 and if one of those fails, it's not clear how the failure would propagate.

You have few solutions. Among the many, a possible one is returning a bool and use the following approach from the caller's code:

```cpp
if (!writeVectorOfStringToMatCell(...)) {
    return false;
}
```

Then, the string could be passed as reference.

A longer discussion about the raw pointer. It's a bit a personal fight, but in C++ I try to minimize their usage because maintaining them is very demanding and memory problems or null pointer access is more likely to happen. A pattern that I like a lot and used in the past (can't find where) is wrapping the raw pointer inside a `unique_ptr`, defining also a custom destructor. In this way, the object will own the memory and when it goes out of scope, it get deallocated automatically. Furthermore, if this pointer is a file, the handle could be closed automatically as well. This approach will provide a bullet proof error handling since you don't have to mind about who has to free the memory in case of errors, making the propagation way more straighforward.

devices/HumanStateProvider/HumanStateProvider.cpp
Yeshasvitvs: Fixed in https://github.com/robotology/human-dynamics-estimation/pull/184/commits/be3f54f2629c6bbb61fc5786ae225d1ec88522c6

devices/HumanStateProvider/HumanStateProvider.cpp
Yeshasvitvs: Fixed in https://github.com/robotology/human-dynamics-estimation/pull/184/commits/be3f54f2629c6bbb61fc5786ae225d1ec88522c6

devices/HumanDataCollector/HumanDataCollector.cpp
Yeshasvitvs: Done https://github.com/robotology/human-dynamics-estimation/pull/181/commits/b8de3bf592fbf86f3100ea63ddee03fddd0eb34d

Yeshasvitvs:
> Few comments not really blocking. It would be nice to have the `unique_ptr` change, I foresee memory related problem in the future and this change would provide a great starting point.

@diegoferigo I am not getting it right. Are you referring to the `unique_prt` to `impl` class ?

devices/HumanDataCollector/HumanDataCollector.cpp
Yeshasvitvs: I am not certain if the clang is correctly setup, I will ensure it. For the time being https://github.com/robotology/human-dynamics-estimation/pull/181/commits/9c24441031952724777a1947969b079deff49c14

devices/HumanDataCollector/HumanDataCollector.cpp
Yeshasvitvs: No particular reason, Corrected in https://github.com/robotology/human-dynamics-estimation/pull/181/commits/7613f51dfe14f3376831210feea7290d7447f82e

devices/HumanDataCollector/HumanDataCollector.cpp
Yeshasvitvs: > I'm not fully conviced that you handle the errros properly. I see a lot of Mat_Close and no error propagation from these write* functions to the outside. You call this method sequentally multiple times around line 850 and if one of those fails, it's not clear how the failure would propagate

@diegoferigo the function `writeVectorOfStringToMatCell()` is called only once per data type like `stateJointNames` or `dynamicsJointNames`. The function `writeVectorOfVectorDoubleToMatStruct()` is called only once to fill the mat file with human data stored in the buffers inside `run()`. Both these methods are called in `detach()` call which is done on termination. So, I do not see the need for error propagation from these methods to `detach()` function. At the moment, it is still a preliminary implementation and it will evolve in the future as https://github.com/robotology/human-dynamics-estimation/issues/179#issuecomment-593951539. Given this possible road map, I would like to proceed with this PR and consider the aspects of raw pointer handling (which is mainly related to matio variables) and error propagation (while logging the data through parallel update of buffer variables) as todo items related to matio logging.

Yeshasvitvs: @diegoferigo https://github.com/robotology/human-dynamics-estimation/pull/181/commits/e1ae3a1d40b8d4ce3dddc0f7f0026307021bf4a4

Yeshasvitvs:
Thanks @diegoferigo @lrapetti 
Merging.

PR #188, opened 2020-02-24
Bug fix: Correct joint data handling from encoder interface in RobotPositionController

Yeshasvitvs:
This PR addresses the crash of `RobotStatePublisher` while using the `PositionDirect` control mode. The following main changes are made:

- Correct the size of the buffer vector to read the joints data from the robot control boards
- Change the detach action level from `5` to `4`, to ensure the correct order of  detaching the devices while terminating `yarprobotinterface`

### DIFF
--- a/conf/xml/RobotStateProvider_iCub.xml
+++ b/conf/xml/RobotStateProvider_iCub.xml
@@ -160,7 +160,7 @@
                 <elem name="RobotPositionController">RobotStateProvider</elem>
             </paramlist>
         </action>
-        <action phase="shutdown" level="5" type="detach"/>
+        <action phase="shutdown" level="4" type="detach"/>
     </device-->
 
 </robot>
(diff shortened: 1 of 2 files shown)


Yeshasvitvs:
@diegoferigo can you please check the failure with Codacy. It highlights "Seems like the devel is not being analyzed. To review this pull request enable the branch in the repository settings."

lrapetti:
> @diegoferigo can you please check the failure with Codacy. It highlights "Seems like the devel is not being analyzed. To review this pull request enable the branch in the repository settings."

I have activated the analysis for `devel` and tried to trigger the analysis. Let's see if it works

Yeshasvitvs:
@lrapetti It looks like the same problem again. If you agree I will proceed with the PR.

lrapetti:
> @lrapetti It looks like the same problem again. If you agree I will proceed with the PR.

yes, we can proceed merging

Issue #192, opened 2020-03-06
Update HumanStateProvider with Rate of Change of Momentum 


Yeshasvitvs:
The current implementation of [computing the rate of change of momentum](https://github.com/robotology/human-dynamics-estimation/blob/feature/visualize-berdy-estimated-wrench/devices/HumanStateProvider/HumanStateProvider.cpp#L1468) is not clean and moreover, the variable names used are misleading. So, in this issue we will track the details related to improving the code pertaining to the rate of change of momentum computation in `HumanStateProvider` device.

Yeshasvitvs:
Consider the reference frames as highlighted in the figure below. The inertial frame of reference is denoted by $\mathcal{I}$, the frame associated with the floating base of the mechanical system is dented as $\mathcal{B}$. The frame associated with the $i$-th link is denoted by $\mathcal{L}_i$. The `centroidal` frame is a unique coordinate frame with its origin at the center of mass of the system and its orientation similar to the orientation of the inertial frame of reference. The centroidal frame is denoted by $\mathcal{G},[\mathcal{I}]$. The gravity vector expressed in the inertial frame of reference is denoted as $\mathbf{g} = \begin{bmatrix} 0 & 0 & -9.81 \end{bmatrix}^\top$

![sot-berdy](https://user-images.githubusercontent.com/6505998/75766724-e38ac000-5d41-11ea-82ad-317882d40ac7.png)




Coming to the case of the rate of change of momentum expressed in centroidal frame, we can separate the linear and angular parts as following,

[...]

PR #193, opened 2020-03-06
Update rate of change of momentum handling 

Yeshasvitvs:
This PR updated the computation of rate of change of momentum according to https://github.com/robotology/human-dynamics-estimation/issues/192#issuecomment-595879160 and its handling through `IHumanState` interface for other devices


claudia-lat:
Please @Yeshasvitvs drop a comment ad eventually evaluate if closing this issue.

Issue #195, opened 2020-05-14
Fix base transform and velocity variable initialization 


Yeshasvitvs:
The base variables [baseTransformSolution and baseVelocitySolution](https://github.com/robotology/human-dynamics-estimation/blob/devel/devices/HumanStateProvider/HumanStateProvider.cpp#L172-L173)  are not initialized correctly.

This issue deals with fixing their initialization

Yeshasvitvs:
A PR https://github.com/robotology/human-dynamics-estimation/pull/196 is open to address this issue

Yeshasvitvs:
PR #196 merged. Closing.

PR #196, opened 2020-05-14
Correct initialization of base transform and velocity variables

Yeshasvitvs:
This patch addresses https://github.com/robotology/human-dynamics-estimation/issues/195


lrapetti:
Looks good

Yeshasvitvs:
Thanks. Merging.

PR #201, opened 2020-07-22
Fix secondary calibration for whole-body

lrapetti:
This PR fix the secondary calibration when used with the whole-body (sending `calibrate` rpc command).


lrapetti:
> LGTM! if you can, please add a screen shot of the terminal if it was showing any error message.

Actually there was no error message but nothing was happening when applying the calibration since the vector with the links to calibrate was empty (it was passed by value instead of reference).

PR #203, opened 2020-07-22
add yaprmanager application for robot xsens retargeting on the pole

kouroshD:
With this PR, we add the yarpmanager application for the robot retargeting with the robot on pole.
We also add the joint limits for the `hip_roll` joints so that it does not touch the pole.

PR #204, opened 2020-07-22
Add humanstate remapper

lrapetti:
This PR introduce the IHumanState remapper device that enable separating the HumanStateProvider from a consumer device (e.g. `RobotPositionController`).


lrapetti:
To be tested in simulation before testing with the robot

lrapetti:
CC @kouroshD in this way we can separate the visualization from the moment in which the controller is started.

Yeshasvitvs:
@lrapetti can you please open an issue and give a bit more background for the need of this new remapper device just to keep track of the changes/features

Yeshasvitvs:
What is the reason behind the choice `IHumanStateRemapper` ? I believe the name `HumanStateRemapper` is more meaningful.

lrapetti:
> @lrapetti can you please open an issue and give a bit more background for the need of this new remapper device just to keep track of the changes/features

sure, I will open it

> What is the reason behind the choice of the name `IHumanStateRemapper` ? I believe the name `HumanStateRemapper` is more meaningful.

I was trying to keep same nomenclature used in [`wearables`](https://github.com/robotology/wearables/)
e.g.
interface: `IWear` -> `IWearRemapper`


Anyway I think that also @diegoferigo might be interested in this discussion.

Yeshasvitvs:
If I am not wrong, `iWear` was the name we were thinking before `iFeel`.

diegoferigo:
For historical reasons we use the pattern `I*` for virtual classes. For example `IWear` and `ISensor`, the are not concrete and they are there to create concrete classes that inherit from them. These concrete classes do not have the `I` prefix.

This being said, if the remapper is a concrete class, to be consistent probably its name should not have the prefix.

lrapetti:
> For historical reasons we use the pattern `I*` for virtual classes. For example `IWear` and `ISensor`, the are not concrete and they are there to create concrete classes that inherit from them. These concrete classes do not have the `I` prefix.

Yes, in this case we are doing the same for the [`IHumanState`](https://github.com/robotology/human-dynamics-estimation/blob/master/interfaces/IHumanState/IHumanState.h) interface class.

> This being said, if the remapper is a concrete class, to be consistent probably its name should not have the prefix.

In this case the remapper class is a concrete class, so in principle non need for `I`. I gave that name since it was the `remapper` for `IHumanState`, so `IHumanStateRemapper`. Analogously I though we had the `Remapper` for `IWear`, so `IWearRemapper`.

Anyway I agree just calling it `HumanStateRemapper` sounds also smooth to me.

lrapetti:
> @lrapetti can you please open an issue and give a bit more background for the need of this new remapper device just to keep track of the changes/features

Issue opened at https://github.com/robotology/human-dynamics-estimation/issues/206

lrapetti:
> Said so, I agree just calling it `HumanStateRemapper` sounds also smooth to me as well.

Changed to `HumanStateRemapper`.

In addiction, I have add the yarpmanager application for the retargeting demo (`RobotPositionController_iCub.xml`), where the `StateProvider` and `RobotPositionController` are separate into two different `yarprobotinterface`

lrapetti:
> In addiction, I have add the yarpmanager application for the retargeting demo (`RobotPositionController_iCub.xml`), where the `StateProvider` and `RobotPositionController` are separate into two different `yarprobotinterface`

I notice this was already done by @kouroshD with https://github.com/robotology/human-dynamics-estimation/pull/203, so I have update the application he created and removed the one I added.

claudia-lat:
@lrapetti any update on this?

lrapetti:
> @lrapetti any update on this?

CI was failing in Ubuntu, trying to fix with https://github.com/robotology/human-dynamics-estimation/pull/204/commits/beb4b310c60b7fe0b6876c75b8aeb45bd5a93286, I'll check the CI output if the error is fixed.

lrapetti:
> CI was failing in Ubuntu, trying to fix with [beb4b31](https://github.com/robotology/human-dynamics-estimation/commit/beb4b310c60b7fe0b6876c75b8aeb45bd5a93286), I'll check the CI output if the error is fixed.

Compilation error was fixed, waiting for reviews.

DanielePucci:
CC @Yeshasvitvs @diegoferigo @kouroshD

conf/xml/Human_Dynamics_from_remapper.xml
```diff
@@ -0,0 +1,187 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE robot PUBLIC "-//YARP//DTD yarprobotinterface 3.0//EN" "http://www.yarp.it/DTD/yarprobotinterfaceV3.0.dtd">
+<robot name="Human-HDE" build=0 portprefix="">
+
+    <device type="ihuman_state_remapper" name="HumanStateRemapper">
```
Yeshasvitvs: @lrapetti are you sure we need this configuration file? and has this been tested ?

I understand the idea we can simply use `HumanStateRemapper` in place of running `HumanStateProvider` device in the same configuration file. However, as we are currently using all the main devices in a single configuration file [Human.xml](https://github.com/robotology/human-dynamics-estimation/blob/devel/conf/xml/Human.xml), we can drop this file as maintaining multiple configuration files that are almost always similar is a bit hard. Instead, we can update [Human.xml](https://github.com/robotology/human-dynamics-estimation/blob/devel/conf/xml/Human.xml) with a comment that says `HumanStateRemapper` device also be used without having to run `HumanStateProvider` device.

devices/HumanDynamicsEstimator/HumanDynamicsEstimator.cpp
```diff
@@ -1195,8 +1195,8 @@ bool HumanDynamicsEstimator::attach(yarp::dev::PolyDriver* poly)
 
     // Get the device name from the driver
     const std::string deviceName = poly->getValue("device").asString();
-
-    if (deviceName == "human_state_provider" || deviceName == "xsens_human_state_provider") {
```
Yeshasvitvs, line 1199: Do you still see the utility of [XsensHumanStateProvider](https://github.com/robotology/human-dynamics-estimation/tree/devel/devices/XsensHumanStateProvider) device. If not, we can remove that device and clean up. What do you think ?

remappers/IHumanStateRemapper/IHumanStateRemapper.h
```diff
@@ -0,0 +1,71 @@
...
+        class IHumanStateRemapper;
+    } // namespace devices
+} // namespace hde
+
+class hde::devices::IHumanStateRemapper final
+    : public yarp::dev::DeviceDriver
+    , public hde::interfaces::IHumanState
+    , public yarp::os::TypedReaderCallback<human::HumanState>
+    , public yarp::os::PeriodicThread
+    //, public yarp::dev::IPreciselyTimed
```
Yeshasvitvs: Clean up

Yeshasvitvs (changes requested):

devices/HumanDynamicsEstimator/HumanDynamicsEstimator.cpp
lrapetti: yes, I think that device was never properly in place, I will proceed removing it.

conf/xml/Human_Dynamics_from_remapper.xml
lrapetti: This was a configuration  file I was using for testing the device, and it is actually "per se" useless. I agree regarding the problem of maintaining multiple files, and I will remove it adding the comment to  `Human.xml`.

remappers/IHumanStateRemapper/IHumanStateRemapper.h
lrapetti: Done in https://github.com/robotology/human-dynamics-estimation/pull/204/commits/2b4fafe9fe3e722b8719b7613bc4cc6d4cbf181b

devices/HumanDynamicsEstimator/HumanDynamicsEstimator.cpp
lrapetti: Done https://github.com/robotology/human-dynamics-estimation/pull/204/commits/c5d70882ea8cb8831a86c7d3c94f333fa9d2ee52

conf/xml/Human_Dynamics_from_remapper.xml
lrapetti: Done https://github.com/robotology/human-dynamics-estimation/pull/204/commits/f294d7cbc5fac1d8300d7979f6cafe89d9322f6b

Yeshasvitvs (approved):
Thanks @lrapetti

Issue #88, opened 2019-01-31
Implement HumanControlBoard device to send human data to simulink


Yeshasvitvs:
Currently [hde-readonly-controlboard](https://github.com/Yeshasvitvs/human-dynamics-estimation/tree/hde_interface/hde-readonly-controlboard) is not compatible with new HDEv2. In this issue we track the details related to updating it to be compatible with taking joint position, velocity and acceleration data coming from `HumanStateProvider` and joint torque data coming from `HumanDynamicsEstimator` to pass the data to matlab simulink

@lrapetti

Yeshasvitvs:
Not necessarily. The first one was designed as a module to work with modules but subscribing to the output ports streaming joint data in thrift format. Now I want to change it to a device called ‘HumanControlBorad’ that will read joint position, velocity through ‘IHumanState’ interface and the subscribe to joint torques output port from ‘HumanDynamicsWrapper’. Later when we implement ‘HumanDynamicsWrapper’ to expose data through ‘IHumanDynamics’ we can update from the logic of reading joint torques port to reading ‘IHumanDynamics’ interface.

diegoferigo:
If I understood what @Yeshasvitvs meant, the new `HumanControlBoard` will have the support of attaching to `IHumanState` and `IHumanDynamics` interfaces. This means that it will run within the same `yarprobotinterface` process (that means same xml file).

This conforms to the architectural choice of HDE v2. As we discussed for all the HDE v2 devices, as soon as we need network transparency we can develop the pairs wrapper-client for each interface. This would allow to run `HumanControlBoard` on a different process / different machine.

Yeshasvitvs:
[HumanControlBoard](https://github.com/Yeshasvitvs/human-dynamics-estimation/tree/feature/humanstateprovider_multi_thread_ik/devices/HumanControlBoard) is added to devices. I implemented only the core methods to access data in matlab simulink through controlboardremapper. Accordingly, we get joint positions and joint velocities from `IHumanState` interface and joint torques from `IHumanDynamics` interace. Ideally, joint accelerations should also be available through `IHumanDynamics` interface but currently they are not set in `HumanDynamicsEstimator`. So, I am setting them to zero in `HumanControlBoard` device.

[...]

PR #213, opened 2020-10-07
Add IControlLimits interface to HumanControlBoard

lrapetti:
Currently the `IControlLimits` interface is not exposed by the `HumanControlBoard`.

The objective of the `HumanControlBoard` is to expose the same `YARP` interfaces that are usually exposed by the ControlBoard of a robot. One of the main purpose is to access the human data in Simulink using WB-Toolbox (see https://github.com/robotology/human-dynamics-estimation/issues/88). 

Among those interfaces there is the `IControlLimits` interface. Currently this interface is not implemented in the `HumanControlBoard`, indeed using the `GetLimits` block (from WB-toolbox) cause an error. In same cases, it can be useful to be able to run this block even with the `HumanControlBoard` (e.g. switching from robot to human case), and so I decided to implement this interface.

Since we don't have such limits for the human, I am setting both upper and lower limits to zero and returning a `Warning` message. Eventually in the future we might add some limits defined accordingly to antropometric measurements as done in the human urdfs in `https://github.com/robotology/human-gazebo`.

### DIFF
--- a/devices/HumanControlBoard/HumanControlBoard.h
+++ b/devices/HumanControlBoard/HumanControlBoard.h
@@ -17,6 +17,7 @@
 #include <yarp/dev/IVelocityControl.h>
 #include <yarp/dev/Wrapper.h>
 #include <yarp/os/PeriodicThread.h>
+#include <yarp/dev/IControlLimits.h>
 
 #include <memory>
 
@@ -36,6 +37,7 @@ class hde::devices::HumanControlBoard final
     , public yarp::dev::IPositionControl
     , public yarp::dev::IVelocityControl
     , public yarp::dev::ITorqueControl
+    , public yarp::dev::IControlLimits
 {
 private:
     class impl;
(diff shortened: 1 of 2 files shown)


diegoferigo (approved):

devices/HumanControlBoard/HumanControlBoard.h
```diff
@@ -130,6 +132,12 @@ class hde::devices::HumanControlBoard final
     bool getTorqueRanges(double* min, double* max) override { return false; }
     bool getMotorTorqueParams(int j, yarp::dev::MotorTorqueParameters* params) override { return false; }
     bool setMotorTorqueParams(int j, const yarp::dev::MotorTorqueParameters params) override { return false; }
+
+    // IControlLimits
+    bool setLimits(int axis, double min, double max) override  { return false; }
```
Yeshasvitvs, line 137: @lrapetti May be a warning message here to indicate setting of limits is not allowed will be indicative

Yeshasvitvs:
@lrapetti if it is not too much work, please add some screenshots indicating the implemented methods in action i.e. getting the values in simulink.

Yeshasvitvs (approved):

lrapetti:
> @lrapetti if it is not too much work, please add some screenshots indicating the implemented methods in action i.e. getting the values in simulink.

Here you can see a Simulink model where we are reading the limits from a `HumanControlBoard` and you can see all the values are set to zero
limit_zero

devices/HumanControlBoard/HumanControlBoard.h
lrapetti: Currently we have no error message for all the `set` methods but we just return `false`. Do you think we should add this kind of message for all the methods?

lrapetti:
I notice CI in MacOS is failing when doing `brew install eigen`, I will try to investigate further but I would proceed merging the PR in the meanwhile.

devices/HumanControlBoard/HumanControlBoard.h
Yeshasvitvs: Don't bother with it for this PR

lrapetti:
Thanks @Yeshasvitvs

PR #216, opened 2020-10-19
Fix macOS CI

lrapetti:
As discussed in https://github.com/robotology/human-dynamics-estimation/issues/214


lrapetti:
The modification seems to solve the problem in the MacOS CI

traversaro (approved):

Issue #212, opened 2020-08-24
Update yarpmanager retargeting application with all possible module dependencies to ensure correct Rviz model visualization 


yeshasvitirupachuri:
Recently we observed that Rviz model visulization was failing when running the [retargeting demo](https://github.com/robotology/human-dynamics-estimation/blob/master/conf/xml/applications/XsensRetargetingVisualization.xml) yarpmanager application. One of the action points to address this issue is to ensure that all the dependencies between different modules are correctly placed, so that before running Rviz model visualization all the data needed for visualization is accessible to Rviz. In this issue will track the details related to adding the module dependencies. 

CC @S-Dafarra @lrapetti @kouroshD

lrapetti:
> Any update @lrapetti? Otherwise we can close it.

Dependencies are still missing (see https://github.com/robotology/human-dynamics-estimation/blob/master/conf/xml/applications/HumanDynamicsEstimation-HumanDynamics.xml#L46-L52), I will try to address this task.

lrapetti:
`yarpmanager` application for the retargeting is now working and the Rviz visualizer has been removed (see https://github.com/robotology/human-dynamics-estimation/blob/master/conf/xml/applications/HDERetargeting_iCub2_5.xml). I think we can keep using it as it is and close it.

PR #220, opened 2020-11-25
Clean configuration files and fix yarpmanager applications

### LINKED ISSUE #116: Update & Test HDE yarpmanager application
Following the creation of `HDE` yarpmanager application https://github.com/dic-iit/component_andy/issues/192 we have to update all the yarpmanager applications and test them and see if all the components are launched correctly through yarpmanager.

CC @lrapetti

lrapetti:
With this PR I am taking care of:
- removing some old deprecated configuration files
- cleaning the Cmake install functions to avoid hardcoding files and avoiding them to be maintained together with the files
- fixing the yarpmanager applications

After this PR, the following Applications are maintained:
- **Human Dynamics** (`HumanDynamicsEstimation-HumanDynamics.xml`): Running Dynamics Estimation with visualization 
- **Human Kinematics** (`HumanDynamicsEstimation-HumanKinematics.xml`): Running Kinematic Estimation only, with visualization 
- **Human dumper** (`HumanDynamicsEstimation-HumanDumper.xml`): for logging HDE related data
- **pHRI** (`HumanDynamicsEstimation-pHRI.xml`): old configuration for Human-Robot experiments, I didn't touch (and tested) this application.
- **Retargeting Position Control Pole** (`XsensRetargetingPositionControlPole.xml`): for running human-iCub retargeting and robot position control, with the robot on the pole.
- **Retargeting visualization** (`XsensRetargetingVisualization.xml`): for visualizing the human-iCub retargeting.

Moreover, there is the application `HumanDynamicsEstimation.xml` that aggregates all the applications above.

Here is a gif with the results running the application for `Human Dynamics` and `Human Kinematics`
HDE application

This PR should close #116 and #212


lrapetti:
Note that, in the iCub retargeting application written by @kouroshD the Host was set to be icub30. For the other applications, I am leaving `localhost` as we want to enable any user to be able to run them on their laptop.

@kouroshD do you think it would make sense to switch also for the iCub applications?

kouroshD:
> @kouroshD do you think it would make sense to switch also for the iCub applications?


@lrapetti  Yes, sure, that sounds good. Please go on with your suggestion.

conf/xml/applications/HumanDynamicsEstimation-HumanKinematics.xml
```diff
@@ -0,0 +1,49 @@
...
+      <port timeout="5.0">/Human/joint_states+@/HumanStatePublisher</port>
+      <port timeout="5.0">/tf+@/HumanStatePublisher</port>
+    </dependencies>
+    <environment>YARP_FORWARD_LOG_ENABLE=1</environment>
+    <description>Run yarprobotstatepublisher with correct human model</description>
+    <node>localhost</node>
+  </module>
+  
+  <!--Rviz-->
+  <module>
```
Yeshasvitvs, line 41: @lrapetti May be adding dependencies of ports opened by HDE devices that stream data through topics for Rviz is helpful.

Yeshasvitvs (approved):
Minor, non-blocking comment for adding dependencies to Rviz module.
Other than that LGTM! Thanks @lrapetti

conf/xml/applications/HumanDynamicsEstimation-HumanKinematics.xml
lrapetti: I checked and RViz but its own can be launched even before the other applications are running (and then once the other applications run you start visualizing the model), so there is no need of waiting for the other applications to stream the data before running RViz.

Indeed for the time being I would proceed merging as it is, and eventually we can add later the dependencies if we see the need for them.

Yeshasvitvs: > I checked and RViz but its own can be launched even before the other applications are running (and then once the other applications run you start visualizing the model)

this is true, but in the past I remember @kouroshD was launching the application and HDE related devices were crashing with Rviz model collapsed. This can be a source of confusion for new users trying to launch the yarp application. 

> Indeed for the time being I would proceed merging as it is, and eventually we can add later the dependencies if we see the need for them.

Agree with this 👍🏾

lrapetti: > I remember @kouroshD was launching the application and HDE related devices were crashing with Rviz model collapsed. This can be a source of confusion for new users trying to launch the yarp application.

sorry, I was not aware of this, then we should investigate it!

PR #223, opened 2020-12-02
Add fixed-base option to HumanStateProvider

lrapetti:
With this PR we add the possibility to run the `HumanStateProvider` with a fixed base link.

In order to set the fixed base it is sufficient to replace the `floatingBaseFrame ` configuration option with `fixedBaseFrame `.

Note that, currently this option would fix both position and orientation of the base. In the future we can think of having the possibility to fix just one of the two.

### DIFF
--- a/devices/HumanStateProvider/HumanStateProvider.cpp
+++ b/devices/HumanStateProvider/HumanStateProvider.cpp
@@ -1002,6 +1002,16 @@ bool HumanStateProvider::open(yarp::os::Searchable& config)
         yInfo() << "CUSTOM CONSTRAINTS are not defined in xml file.";
     }
 
+    // set base velocity constraint to zero if the base is fixed
+    if (pImpl->useFixedBase) {
+        pImpl->baseVelocityLowerLimit.resize(6);
+        pImpl->baseVelocityLowerLimit.zero();
+        pImpl->baseVelocityUpperLimit.resize(6);
+        pImpl->baseVelocityUpperLimit.zero();
+
+        yInfo() << "Using fixed base model, base velocity limits are set to zero";
+    }
+
     // check sizes
     if (pImpl->custom_jointsVelocityLimitsNames.size()
         != pImpl->custom_jointsVelocityLimitsValues.size()) {
--- a/devices/HumanStateProvider/HumanStateProvider.cpp
+++ b/devices/HumanStateProvider/HumanStateProvider.cpp
@@ -239,7 +239,9 @@ public:
 
     SolverIK ikSolver;
 
+    // flags
     bool useDirectBaseMeasurement;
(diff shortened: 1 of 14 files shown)


devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -438,9 +437,19 @@ bool HumanStateProvider::open(yarp::os::Searchable& config)
         return false;
     }
 
-    if (!(config.check("floatingBaseFrame") && config.find("floatingBaseFrame").isList()
-          && config.find("floatingBaseFrame").asList()->size() == 2)) {
-        yError() << LogPrefix << "floatingBaseFrame option not found or not valid";
+    yarp::os::Bottle* floatingBaseFrameList;
```
Yeshasvitvs, line 440: @lrapetti may be using two parameters as following:

- `baseFrame` that is a list of size two
- `fixedBase` that is a bool, which if set to false will be a floating base configuration

will be a clean and concise code. What do you think ?

Yeshasvitvs (changes requested):

Yeshasvitvs:
Ping @lrapetti

devices/HumanStateProvider/HumanStateProvider.cpp
lrapetti: In both the case we should pass the name of the model base frame, indeed we can not use a bool. 

Anyway, I have noticed we were passing this option as a `List` but we are just using the first element of the list (the name of the base frame in the model). Indeed with https://github.com/robotology/human-dynamics-estimation/pull/223/commits/3add0d577519d2594cfe4073b5aaaefdc2fb0c2f I add the possibility to pass this option simply as a string, and I have added a deprecation warning when it is passed as list (even if the list is still supported in order to avoid breaking old configuration files).  I have also changed the name of the variable `floatingBaseFrameList` to `baseFrameName` in order t be more clean.

Yeshasvitvs: I think the choice of using `floatingBaseFrame` as a list https://github.com/robotology/human-dynamics-estimation/blob/devel/conf/xml/Human.xml#L23 was based on the choice of link and joint map information parameters we used https://github.com/robotology/human-dynamics-estimation/blob/devel/conf/xml/Human.xml#L55-L79

Thanks for updating it and keeping back compatibility.

Yeshasvitvs (approved):

lrapetti:
I have rebased the branch on top of latest `devel`, waiting for the checks before merging

Yeshasvitvs:
Ping @lrapetti

lrapetti:
> Ping @lrapetti

Checks are passed, we can merge

PR #237, opened 2021-02-08
Add base calibration

lrapetti:
With this PR, we add the possibility to apply the secondary calibration also to the model base position and orientation.

This calibration procedure works analogously to the `calibrateAll` command, but the difference is that also a secondary calibration is applied to the base position and orientation.


lrapetti:
I would like to run a couple of further test before merging, I will add the video of the outcome in this PR

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -319,7 +322,8 @@ class HumanStateProvider::impl::CmdParser : public yarp::os::PortReader
             if (command.get(0).asString() == "help") {
                 response.addVocab(yarp::os::Vocab::encode("many"));
                 response.addString("The following commands can be used to apply a secondary calibration assuming the subject is in the zero configuration of the model for the calibrated links. \n");
-                response.addString("Enter <calibrateAll> to apply a secondary calibration for all the links \n");
+                response.addString("Enter <calibrateAll> to apply a secondary calibration for all the links using the measured base pose \n");
+                response.addString("Enter <calibrateAllWithBase <baseLink>> to apply a secondary calibration for all the links assuming the <baseLink> to be in the origin \n");
```
Yeshasvitvs, line 326: @lrapetti I am a bit unclear of the usage of this command. Does `calibrateAllWithBase` can be applied to any link to be considered as base link for consideration ? e.g. `calibrateAllWithBase RightHand`.

Or `calibrateAllWithBase` works with the link that is set as the base link. In this case, the input command `calibrateAllWithBase <baseLink>` is redundant, right ?

lrapetti:
Today I have tested the PR again and I have found out that it was not behaving as expected.
Indeed with https://github.com/robotology/human-dynamics-estimation/pull/237/commits/ccd7687fb1cec0ede2e5d75342968fcce8bc7736 the problems should be fixed.

Note that I have changed the new method name to `calibrateAllWithWorld` because the main difference with `calibrateAll` command, it is that we apply also a pre-multiplication calibration matrix which change the world in order to centre it in the desired link (the link given as argument).
This is documented in https://github.com/lrapetti/human-dynamics-estimation/blob/baf97260d984b42b012224da69a37b2fc0ef49a1/devices/HumanStateProvider/HumanStateProvider.cpp#L327.

I will still run some test, and put here the output, before considering the PR ready to be merged.

lrapetti:
The following video shows the different output between the two calibration methods:

- `calibrateAll`

https://user-images.githubusercontent.com/35487806/109328533-d2439900-7859-11eb-9ba9-9f4cfdb50aa7.mov

- 'calibrateAllWithWorld LeftToe'

https://user-images.githubusercontent.com/35487806/109328562-dc659780-7859-11eb-94f7-e1e69e9c4dfd.mov

devices/HumanStateProvider/HumanStateProvider.cpp
lrapetti: This point should be more clear with the updated command discussed in https://github.com/robotology/human-dynamics-estimation/pull/237#issuecomment-786744471:
```
calibrateAllWithWorld <linkName>
```
it calibrate all the links, assuming the `<linkName>` is the world origin frame.

lrapetti:
@Yeshasvitvs @prashanthr05 I have updated the PR and tested the results (https://github.com/robotology/human-dynamics-estimation/pull/237#issuecomment-786758973), I think you can proceed with the reviews.

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -1333,14 +1353,30 @@ void HumanStateProvider::impl::computeSecondaryCalibrationRotationsForChain(cons
...
-            iDynTree::Rotation linkRotationZero = kinDynComputations->getWorldTransform(linkToCalibrateName).getRotation();
-            iDynTree::Rotation secondaryCalibrationRotation = linkTransformMatricesRaw.at(linkToCalibrateName).getRotation().inverse() * linkRotationZero;
+
+            iDynTree::Transform linkTransformZero = kinDynComputations->getWorldTransform(linkToCalibrateName);
+
+            // computing new calibration for orientation
+            iDynTree::Rotation secondaryCalibrationRotation = linkTransformMatricesRaw.at(linkToCalibrateName).getRotation().inverse() * linkTransformZero.getRotation();
+
+            // compute secondary calibration for position
+            iDynTree::Position secondaryCalibrationPosition = linkTransformZero.getPosition() - linkTransformMatricesRaw.at(linkToCalibrateName).getPosition();
```
prashanthr05, line 1363: Just curious, why this would be `p1 - p2` and not be `p1 - (R1' * p2)` ? They're both already in the same coordinate frames?

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -1378,21 +1414,43 @@ bool HumanStateProvider::impl::applyRpcCommand()
...
+        }
+
+        // Select all the links and the joints
+        // add all the links of the model to [linkToCalibrateIndices]
+        linkToCalibrateIndices.resize(kinDynComputations->getNrOfLinks());
+        std::iota(linkToCalibrateIndices.begin(), linkToCalibrateIndices.end(), 0);
+
+        // add all the joints of the model to [jointZeroIndices]
+        jointZeroIndices.resize(kinDynComputations->getNrOfDegreesOfFreedom());
+        std::iota(jointZeroIndices.begin(), jointZeroIndices.end(), 0);
```
prashanthr05, line 1436: Nice, thanks! Didn't know about `iota` before.

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -225,10 +226,12 @@ class HumanStateProvider::impl
 
     // Secondary calibration
     std::unordered_map<std::string, iDynTree::Rotation> secondaryCalibrationRotations;
+    std::unordered_map<std::string, iDynTree::Position> secondaryCalibrationPositions;
+    iDynTree::Transform secondaryCalibrationWorld;
     void eraseSecondaryCalibration(const std::string& linkName);
     void selectChainJointsAndLinksForSecondaryCalibration(const std::string& linkName, const std::string& childLinkName,
                                                   std::vector<iDynTree::JointIndex>& jointZeroIndices, std::vector<iDynTree::LinkIndex>& linkToCalibrateIndices);
-    void computeSecondaryCalibrationRotationsForChain(const std::vector<iDynTree::JointIndex>& jointZeroIndices, const iDynTree::Transform &baseTransform, const std::vector<iDynTree::LinkIndex>& linkToCalibrateIndices);
+    void computeSecondaryCalibrationRotationsForChain(const std::vector<iDynTree::JointIndex>& jointZeroIndices, const iDynTree::Transform &baseTransform, const std::vector<iDynTree::LinkIndex>& linkToCalibrateIndices, const std::string& baseLinkName);
```
prashanthr05, line 234: A curiosity, would it be more meaningful to rename `baseLinkName` to `refLinkForCalibration` ? I suppose this base link is not the same as the base link of the `kindyncomputations` ? We are not actually changing the default base link of the model with this function call right? or did I miss something?

devices/HumanStateProvider/HumanStateProvider.cpp
lrapetti: me neither, I just found it on the web :grimacing:

devices/HumanStateProvider/HumanStateProvider.cpp
lrapetti: Yes, you are right. There is actually no reason for referring at it as baseLink, and it is, in fact, a source of confusion (e.g. https://github.com/robotology/human-dynamics-estimation/pull/237#discussion_r572866358).

I will change it accordingly to your suggestion.

lrapetti: Done https://github.com/robotology/human-dynamics-estimation/pull/237/commits/a7f8724e968dfc7e5adc5b14a15bbd6d81735348

devices/HumanStateProvider/HumanStateProvider.cpp
lrapetti: Yes, I think you are right. Anyway, actually are not using secondary calibration for the position of the links (see https://github.com/lrapetti/human-dynamics-estimation/blob/add-base-calibration/devices/HumanStateProvider/HumanStateProvider.cpp#L1556) since we are just considering the link orientation for the inverse kinematics, and the base offset is applied together with the [secondaryCalibrationWorld](https://github.com/lrapetti/human-dynamics-estimation/blob/add-base-calibration/devices/HumanStateProvider/HumanStateProvider.cpp#L1559) which contains both position and orientation offset of the world.

I will probably proceed directly removing the `secondaryCalibrationPosition`

lrapetti: Removed in https://github.com/robotology/human-dynamics-estimation/pull/237/commits/941ab91c08b2637a19795c7be83761849a1401f2

prashanthr05 (approved):
In general, lgtm. I would wait for @Yeshasvitvs approval, though.

Yeshasvitvs:
Thanks @lrapetti LGTM!

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -1419,10 +1419,10 @@ bool HumanStateProvider::impl::applyRpcCommand()
     }
     case rpcCommand::calibrateAllWithWorld: {
         // Check if the chose baseLink exist in the model
-        std::string baseLinkName = commandPro->baseLinkName;
-        if(!(kinDynComputations->getRobotModel().isFrameNameUsed(baseLinkName)))
+        std::string refLinkForCalibrationName = commandPro->refLinkName;
+        if(!(kinDynComputations->getRobotModel().isFrameNameUsed(refLinkForCalibrationName)))
```
Yeshasvitvs, line 1416: @lrapetti  Just a minor comment non-blocking comment, checking this link name in the model instance, rather than the kindyncomputations instance may be more certain check. 

Check https://github.com/robotology/human-dynamics-estimation/blob/a7f8724e968dfc7e5adc5b14a15bbd6d81735348/devices/HumanStateProvider/HumanStateProvider.cpp#L809

and 

https://github.com/robotology/human-dynamics-estimation/blob/a7f8724e968dfc7e5adc5b14a15bbd6d81735348/devices/HumanStateProvider/HumanStateProvider.cpp#L818

Yeshasvitvs (approved):

devices/HumanStateProvider/HumanStateProvider.cpp
lrapetti: If it is fine to you I would proceed merging as it is

Yeshasvitvs: Sure, go ahead 😉

Issue #240, opened 2021-03-02
Add forces visualization in HumanStateVisualizer


lrapetti:
Since I was working on force visualization with the iDynTree based visualizer, I was thinking of adding it also to the HumanStateVisualizer so that we get closer to the Rviz based visualizer aspect.

lrapetti:
The implementation can be found in [`feature/add-arrow-viz`](https://github.com/robotology/human-dynamics-estimation/tree/feature/add-arrow-viz). The feature is already working:

https://user-images.githubusercontent.com/35487806/109705090-a563ee80-7b97-11eb-822d-e15b218ef776.mov

the only point before opening the PR and merging is that it depends on a fix done by @S-Dafarra (see https://github.com/robotology/idyntree/issues/836#issuecomment-788943524) and not yet merged in master.
As soon as it is merged I will proceed with this feature.

PR #241, opened 2021-03-03
Add force arrows to the HumanStateVisualizer

lrapetti:
This PR close https://github.com/robotology/human-dynamics-estimation/issues/240 by adding the possibility to visualize the force associated to a given link by reading its value from the `HumanWrenchProvider` trough an `analogServer`.

The PR is ready, as shown in https://github.com/robotology/human-dynamics-estimation/issues/240#issuecomment-789165160, but waiting for a fix in `iDynTree` to be merged (https://github.com/robotology/idyntree/pull/838)

### DIFF
--- a/conf/app/HumanStateVisualizer.ini
+++ b/conf/app/HumanStateVisualizer.ini
@@ -3,6 +3,7 @@ name                    HumanStateVisualizer
 # Model Configuration options
 modelURDFName      "humanSubject01_66dof.urdf"
 ignoreMissingLinks  true
+visualizeWrenches   true
 
 # Camera options
 cameraDeltaPosition  (2.0, 0.0, 0.5)
@@ -10,5 +11,10 @@ useFixedCamera       false           # if set to false, the camera follows the m
 fixedCameraTarget    (0.0, 0.0, 0.0) # this option is unused when useFixedCamera is false
 maxVisualizationFPS  65
 
+# Wrench Data Configuration
+forceScalingFactor   0.001
+wrenchSourceLinks    (LeftFoot RightFoot) # link order should reflect WrenchWrapper port data
+
 # Remapper Configuration
 humanStateDataPortName "/HDE/HumanStateWrapper/state:o"
+humanWrenchWrapperPortName "/HDE/HumanWrenchWrapper/wrench:o"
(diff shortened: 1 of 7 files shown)


lrapetti:
> The PR is ready, as shown in [#240 (comment)](https://github.com/robotology/human-dynamics-estimation/issues/240#issuecomment-789165160), but waiting for a fix in `iDynTree` to be merged ([robotology/idyntree#838](https://github.com/robotology/idyntree/pull/838))

The PR on iDynTree was merged, so we can proceed with merging also this PR.

Note that currently, the new feature will work properly only on `master`, but still it compiles and runs with previous tags (e.g. the [CI tag `3.0.0`](https://github.com/robotology/human-dynamics-estimation/blob/master/.github/workflows/ci.yml#L19)) but the force vectors will be positioned in the origin because of the bug described in https://github.com/robotology/idyntree/issues/835

S-Dafarra (approved):
Looks good to me. I would just argue that reading two wrenches simply as 12 numbers without any other indication is a bit risky in terms of portability. You need to make sure that the configuration file is always aligned with what provides you those forces, but I think this is not the aim of this PR in any case.

lrapetti:
> Looks good to me. I would just argue that reading two wrenches simply as 12 numbers without any other indication is a bit risky in terms of portability.

I completely agree. Unfortunately that is how the force values are streamed by the [`HumanWrenchProvider`](https://github.com/robotology/human-dynamics-estimation/blob/master/devices/HumanWrenchProvider/HumanWrenchProvider.cpp), in order to have a better solution we should first improve the way those data are exposed by that device.

lrapetti:
> > Looks good to me. I would just argue that reading two wrenches simply as 12 numbers without any other indication is a bit risky in terms of portability.
> 
> I completely agree. Unfortunately that is how the force values are currently streamed by the [`HumanWrenchProvider`](https://github.com/robotology/human-dynamics-estimation/blob/master/devices/HumanWrenchProvider/HumanWrenchProvider.cpp), in order to have a better solution we should first improve the way those data are exposed by that device.

I also considered reading direcly the wrench values for the shoes (e.g. using IWear interface), but in that case we should have passed to the visualizer some parameters ([shoes frame position and orientation](https://github.com/robotology/human-dynamics-estimation/blob/master/conf/xml/Human.xml#L210-L213)), which would have anyway create confusion in configuration files.

PR #252, opened 2021-05-26
Add the possibility to use the HumanControlBoard with just kineamtics or dynamics

lrapetti:
Currently, the `HumanControlBoard`, which is the device that exposes the output of the HDE as a robot controlboard, is working only if both `HumanStateProvider` and `HumanDynamicsEstimator` are running.

With this PR, I add the possibility of running it, even with just one of the two. This is possible by changing the [network elements](https://github.com/robotology/human-dynamics-estimation/blob/master/conf/xml/Human.xml#L291-L292). To the desired one. Data related to the other device are fixed to be zero.

I have tested the device in combination with the [`WB-toolbox`](https://github.com/robotology/wb-toolbox), after the changes and it is working fine.


lrapetti:
In the PR I am also removing a left-over deprecated device (`RobotPositionController-old.cpp`)

devices/HumanControlBoard/HumanControlBoard.cpp
```diff
@@ -132,36 +134,60 @@ bool HumanControlBoard::open(yarp::os::Searchable& config)
...
+    {
+        yError() << LogPrefix << "No interface attached to the device.";
+        askToStop();
+        return;
+    }
 
-    std::vector<double> jointPositionsInterface = pImpl->iHumanState->getJointPositions();
-    std::vector<double> jointVelocitiesInterface = pImpl->iHumanState->getJointVelocities();
+    // Initialize buffers for measurements
+    std::vector<double> jointPositionsInterface(pImpl->nJoints, 0.0);
```
Yeshasvitvs, line 161: Minor: The variable ending of `Interface` is a bit misleading. May be we can drop it as : `jointPositionsInterface` -------> `jointPositions`

Also for other variables

Yeshasvitvs: or a suffix as `Data`

Yeshasvitvs (approved):
Thanks @lrapetti LGTM. Just a minor comment about variable name.
It would also be nice if there is any video or screenshot for future reference

devices/HumanControlBoard/HumanControlBoard.cpp
lrapetti: Yes, I was also not very convinced by that variable naming. I have changed to `jointXXXData`

Yeshasvitvs:
Thanks @lrapetti 
Merged.

PR #254, opened 2021-06-01
Add Human State Provider new features

lrapetti:
This PR introduces some new features for the `HumanStateProvider`:
- Add the possibility to run fixed-based inverse kineamtics, using a sensor as fixed base, and computing the other sensors relative pose (https://github.com/robotology/human-dynamics-estimation/commit/1381f42de3783913aa859d65b877ad1bd699fe63)
- Add optional `jointList` parameter for `HumanStateProvider` device: Add the possibility to use a reduced model by passing list of links from the configuration (https://github.com/robotology/human-dynamics-estimation/commit/5d6dde63c3e2cd998d7f78da53c93866db75c740).
- Add the possibility to pass a fixed rotation for a sensor measurement trough configuration files (https://github.com/robotology/human-dynamics-estimation/commit/a4b383cc320df60980698ac2b7269bacef42ba0b)

### DIFF
--- a/devices/HumanStateProvider/HumanStateProvider.cpp
+++ b/devices/HumanStateProvider/HumanStateProvider.cpp
@@ -163,6 +163,8 @@ public:
     iDynTree::Model humanModel;
     FloatingBaseName floatingBaseFrame;
 
+    std::vector<std::string> jointList;
+
     std::vector<SegmentInfo> segments;
     std::vector<LinkPairInfo> linkPairs;
 
@@ -442,6 +444,19 @@ bool HumanStateProvider::open(yarp::os::Searchable& config)
     if (!(config.check("useXsensJointsAngles") && config.find("useXsensJointsAngles").isBool())) {
         yError() << LogPrefix << "useXsensJointsAngles option not found or not valid";
         return false;
+    }
+
+    if (!(config.check("jointList") && config.find("jointList").isList())) {
+        yInfo() << LogPrefix << "jointList option not found or not valid, all the model joints are selected.";
+        pImpl->jointList.clear();
+    }
+    else
+    {
+        auto jointListBottle = config.find("jointList").asList();
+        for (size_t it = 0; it < jointListBottle->size(); it++)
+        {
+            pImpl->jointList.push_back(jointListBottle->get(it).asString());
+        }
     }
 
     std::string baseFrameName;
@@ -789,10 +804,21 @@ bool HumanStateProvider::open(yarp::os::Searchable& config)
     }
 
     iDynTree::ModelLoader modelLoader;
-    if (!modelLoader.loadModelFromFile(urdfFilePath) || !modelLoader.isValid()) {
+    if ( pImpl->jointList.empty())
+    {
+        if (!modelLoader.loadModelFromFile(urdfFilePath) || !modelLoader.isValid()) {
         yError() << LogPrefix << "Failed to load model" << urdfFilePath;
         return false;
+        }
     }
+    else
+    {
+        if (!modelLoader.loadReducedModelFromFile(urdfFilePath, pImpl->jointList) || !modelLoader.isValid()) {
+        yError() << LogPrefix << "Failed to load model" << urdfFilePath;
+        return false;
+        }
+    }
+    
     yInfo() << LogPrefix << "----------------------------------------" << modelLoader.isValid();
     yInfo() << LogPrefix << modelLoader.model().toString();
     yInfo() << LogPrefix << modelLoader.model().getNrOfLinks()
--- a/devices/HumanStateProvider/HumanStateProvider.cpp
+++ b/devices/HumanStateProvider/HumanStateProvider.cpp
@@ -228,6 +228,7 @@ public:
 
     // Secondary calibration
     std::unordered_map<std::string, iDynTree::Rotation> secondaryCalibrationRotations;
+    std::unordered_map<std::string, iDynTree::Rotation> fixedSensorRotation;
     iDynTree::Transform secondaryCalibrationWorld;
     void eraseSecondaryCalibration(const std::string& linkName);
     void selectChainJointsAndLinksForSecondaryCalibration(const std::string& linkName, const std::string& childLinkName,
@@ -255,6 +256,7 @@ public:
     bool getJointAnglesFromInputData(iDynTree::VectorDynSize& jointAngles);
     bool getLinkTransformFromInputData(std::unordered_map<std::string, iDynTree::Transform>& t);
     bool computeRelativeTransformForInputData(std::unordered_map<std::string, iDynTree::Transform>& t);
+    bool applyFixedTransformForInputData(std::unordered_map<std::string, iDynTree::Transform>& t);
     bool getLinkVelocityFromInputData(std::unordered_map<std::string, iDynTree::Twist>& t);
 
     // calibrate data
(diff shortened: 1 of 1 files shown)


Yeshasvitvs:
> Add the possibility to use a reduced model by passing list of links from the configuration (5d6dde6)

If I am not wrong is the joint list we pass that handles the link to be loaded in the model right? If so, please update the main comment. Thanks.

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -443,6 +448,19 @@ bool HumanStateProvider::open(yarp::os::Searchable& config)
         return false;
     }
 
+    if (!(config.check("jointList") && config.find("jointList").isList())) {
```
Yeshasvitvs, line 451: May be adding in the main comment a sentence as below will be easy to track the changes.

> Add optional `jointList` parameter for `HumanStateProvider` device

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -489,6 +507,35 @@ bool HumanStateProvider::open(yarp::os::Searchable& config)
         }
     }
 
+
+    yarp::os::Bottle& fixedRotationGroup = config.findGroup("FIXED_SENSOR_ROTATION");
+    if (!fixedRotationGroup.isNull()) {
+        for (size_t i = 1; i < fixedRotationGroup.size(); ++i) {
```
Yeshasvitvs, line 513: May be a debug info showing the use of `FixedRotationGroup`  config parameter will be useful

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -489,6 +507,35 @@ bool HumanStateProvider::open(yarp::os::Searchable& config)
...
+            if (!(fixedRotationGroup.get(i).isList() && fixedRotationGroup.get(i).asList()->size() == 2)) {
+                yError() << LogPrefix
+                        << "Childs of FIXED_SENSOR_ROTATION must be lists of 2 elements";
+                return false;
+            }
+            yarp::os::Bottle* list = fixedRotationGroup.get(i).asList();
+            std::string key = list->get(0).asString();
+            yarp::os::Bottle* listContent = list->get(1).asList();
+
+            if (!(    (listContent->size() == 9)
```
Yeshasvitvs, line 523: This is a bit vague to follow without any additional comments about the list of values passed in the configuration

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -1533,6 +1629,25 @@ bool HumanStateProvider::impl::getLinkTransformFromInputData(
...
+{
+    // if the there is a measurement for the floating base,
+    // use the relative transform for the other sensors measurement
+    if (transforms.find(floatingBaseFrame) != transforms.end())
+    {
+        iDynTree::Rotation baseFrameRotationInvese = transforms[floatingBaseFrame].getRotation().inverse();
+        for (const auto& linkMapEntry : wearableStorage.modelToWearable_LinkName) {
+            const ModelLinkName& modelLinkName = linkMapEntry.first;
+
+            iDynTree::Rotation rotation = baseFrameRotationInvese * transforms[modelLinkName].getRotation();
```
Yeshasvitvs, line 1643: Typo `baseFrameRotationInvese` ----------> `baseFrameRotationInverse`

Yeshasvitvs (changes requested):
@lrapetti Thanks for the PR
Just some minor comments.

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -443,6 +448,19 @@ bool HumanStateProvider::open(yarp::os::Searchable& config)
         return false;
     }
 
+    if (!(config.check("jointList") && config.find("jointList").isList())) {
+        yInfo() << LogPrefix << "jointList option not found or not valid, all the model joints are selected.";
+        pImpl->jointList.clear();
```
prashanthr05, line 453: Do you think in this case where the joints were not specified, it would be helpful to return the joints list used by the loaded model back to the user? This can help reduce uncertainties for logging and testing with some data.

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -1533,6 +1629,25 @@ bool HumanStateProvider::impl::getLinkTransformFromInputData(
     return true;
 }
 
+bool HumanStateProvider::impl::computeRelativeTransformForInputData(
```
prashanthr05, line 1632: This function is not clear to me in what it does. It takes pose of links in inertial frames and converts it to pose in base frame. Why are we doing this? and why are we modifying the same container that is storing the former into latter?

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -1558,6 +1673,27 @@ bool HumanStateProvider::impl::applySecondaryCalibration(
     return true;
 }
 
+bool HumanStateProvider::impl::applyFixedTransformForInputData(std::unordered_map<std::string, iDynTree::Transform> &transforms_in)
```
prashanthr05, line 1676: what do you think about renaming this to `applyFixedRightRotationForInputTransformData()` ?

devices/HumanStateProvider/HumanStateProvider.cpp
lrapetti: R

devices/HumanStateProvider/HumanStateProvider.cpp
lrapetti: I think some information are given below when the model is loaded. This device is already quite verbose, so I would avoid adding further messages (I would rather try to remove some of them in the future).

devices/HumanStateProvider/HumanStateProvider.cpp
lrapetti: This is done in case the following two conditions are verified:
- fixed base is used
- a sensor measurement associated with the base is present

This is the case for example, in which we have a sensor attached to the base frame, but we want to solve the fixed-base inverse kinematics. In this case, in fact, we have to consider only the relative rotation between the other sensors and the base sensors. Please let me know if that is clear.

I will also change it in order to have input-output transforms.

devices/HumanStateProvider/HumanStateProvider.cpp
lrapetti: Sure, I can do it. I would also like to rename all the methods/commands adding `right`/`left` rotation keyworkds, but probably in a future PR.

devices/HumanStateProvider/HumanStateProvider.cpp
lrapetti: > I will also change it in order to have input-output transforms.

Actually, I noticed we modify the same container also in other points. Do you think that is kind of bad, or is it fine? Since some of those methods are optional, it is kind of convient to use always the same container since we avoid to instantiate multiple variable.

devices/HumanStateProvider/HumanStateProvider.cpp
lrapetti: thanks! Fixed.

devices/HumanStateProvider/HumanStateProvider.cpp
lrapetti: I added a comment, at this stage we are passing 9 values since we pass the full transform matrix in the configuration file.

devices/HumanStateProvider/HumanStateProvider.cpp
lrapetti: I add a log message with the links on which secondary calibration is applied

devices/HumanStateProvider/HumanStateProvider.cpp
prashanthr05: > This is done in case the following two conditions are verified:
> - fixed base is used
> - a sensor measurement associated with the base is present
> This is the case for example, in which we have a sensor attached to the base frame, but we want to solve the fixed-base inverse kinematics. In this case, in fact, we have to consider only the relative rotation between the other sensors and the base sensors. 

maybe we could add what you wrote here as comments/documentation. That should be helpful for the future.

> Do you think that is kind of bad, or is it fine? Since some of those methods are optional, it is kind of convient to use always the same container since we avoid to instantiate multiple variable.

It might seem convenient at the moment, but there is a possibility that it might come back to bite us in the future, in  case we engage in some rigorous debugging. That's my only concern. 
I see your point as well because In the other case, I think it's harder to track two different containers.

if you believe this will not become a big issue for the future, then we can proceed as it is.

lrapetti: Ok, yes, for the time being, I would leave it as it is.

Meanwhile, I have added a comment that documents what I wrote above on the usage of this method.

lrapetti:
Hi @prashanthr05 @Yeshasvitvs, let me know if you have further comments or we can merge

prashanthr05:
@lrapetti  None from my side, thanks!

Yeshasvitvs (approved):
Thanks @lrapetti

PR #257, opened 2021-06-14
Fix joint limits read

lrapetti:
Currently, there is a bug in the joint limits that are passed to the `stateIntegrator`. The error is due to the fact that we are iterating on DOFs but then the output is used for reading the joint.
The issue should be fixed by this PR.

In addition, I am fixing the iCub3 teleoperation model on the right wrist (see https://github.com/robotology/icub-models/issues/93)

### DIFF
--- a/conf/urdfs/teleoperation_iCub_model_V_3.urdf
+++ b/conf/urdfs/teleoperation_iCub_model_V_3.urdf
@@ -864,12 +864,13 @@
       </geometry>
     </collision>
   </link>
-  <joint name="r_wrist_yaw" type="continuous">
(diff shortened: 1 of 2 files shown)


S-Dafarra (approved):

lrapetti:
Thanks @S-Dafarra.

Let's recall that https://github.com/robotology/human-dynamics-estimation/pull/257/commits/2a1fba7d5a7b8c7596141d2c91f141e865474094 is a temporary patch, and we should check again after the bug is fixed on the upstream model (https://github.com/robotology/icub-models/issues/93)

Issue #239, opened 2021-03-01
Add documentation for calibration procedure


lrapetti:
In time we add multiple different procedures for secondary calibration of the kinematics data.
Those are now documented in PRs/issues or in the rpc response (https://github.com/robotology/human-dynamics-estimation/blob/6e3ada1169387b3105d01b2cd97b7d1cb7e1f93b/devices/HumanStateProvider/HumanStateProvider.cpp#L318-L329).

It would be more proper to add a documentation in the `README`

claudia-lat:
Any update @lrapetti ? Otherwise we can close it.

lrapetti:
PR opened https://github.com/robotology/human-dynamics-estimation/pull/264

lrapetti:
Done in #264, we can close it

PR #264, opened 2021-10-15
Add kinematic calibration documentation

lrapetti:
Adding minimum information regarding kinematic calibration (see https://github.com/robotology/human-dynamics-estimation/issues/239#issuecomment-943294948), and improving some formatting.


lrapetti:
Thanks @RiccardoGrieco !

PR #272, opened 2021-11-11
Human State Provider: fix the bug related to the zero output qp solver

kouroshD:
With this PR, I am fixing the problem related to the QP that returns zero values as the solution.
reference issue: https://github.com/robotology/human-dynamics-estimation/issues/167

The format of the files updated are set to clang format.


kouroshD:
Probably, we can add the default solver to QP. what do you say @lrapetti ?!

kouroshD:
I have separated the commit for the clang format updates of the files and the real commit for the bug fix.
Ready to be reviewed @lrapetti

lrapetti:
Thanks, @kouroshD!
I just realized that it would be better not to merge the clang form at this stage since it would increase a lot the complexity for rebasing the child banch on which I have already started working (https://github.com/robotology/human-dynamics-estimation/tree/add-different-target-types).

If for you is ok, I would remove the clang commit and merge this fix.

kouroshD:
@lrapetti  Feel free to proceed as you suggested.

lrapetti:
> @lrapetti Feel free to proceed as you suggested.

Done.
I will merge it in the main PR (https://github.com/robotology/human-dynamics-estimation/pull/265)

Issue #273, opened 2021-11-15
Add the possibility to set different type of targets and different type of sensors in HumanStateProvider


lrapetti:
Currently, the `human-state-provider` device is implemented so that all the measurement targets are added as angular velocity (rotational) targets, and a single linear velocity (position) target is used just for the base frame (see [here](https://github.com/robotology/human-dynamics-estimation/blob/master/devices/HumanStateProvider/HumanStateProvider.cpp#L2501-L2518)). Moreover, only `virtualLink` wearable data source are used.

We should add, instead, the possibility to use multiple linear and angular targets from different wearable data sources.
In order to do so, we could update the [configuration file](https://github.com/robotology/human-dynamics-estimation/blob/master/conf/xml/Human.xml#L55-L79) with a field that specifies the type of target (e.g. position, position-velocity, orientation, orientation-angularVelocity) and then define the targets accordingly.

lrapetti:
Feature merged onto the main development branch `add-different-target-types`, we can close this issue.

Issue #208, opened 2020-07-27
Improve secondary calibration


lrapetti:
Some options of the secondary calibration are currently not working for some model (e.g. the chain tree calibration for the `iCub` model). Those problem should be investigated, and furthermore a general refactoring might be useful.

I'll first try to sum-up what is there, what is working and what not.

lrapetti:
Up to now there are three calibration strategies:
- **whole-body calibration**
  **command:** `calibrate`
  It assumes the subject is assuming the zero configuration $\tilde{s}=s_0=0$ (i.e. the configuration in which the model joints are all zero), while for the base link is assumed being the measured one. At this point the secondary calibration matrix is computed for each link as follow:
  $$
  R_i^{II}=R_i(p_B(t),R_{B}(t),s(t))^TR_i(p_B(t),R_{B}(t),s_0)
  $$
  
- **link calibration**
  **command:** `calibrate <linkName>`
  It assumes that the given link $L$ is in zero configuration w.r.t. its child link. So given the set of joints $\mathcal{J}$ between child and parent link, 
  $$
  \begin{equation}
  \tilde{s}_j=s_j(t) \ \forall j \notin \mathcal{J} \\\\
  \tilde{s}_j=0 \ \ otherwise
  \end{equation}
  $$
  so at this point we can define the secondary calibration matrixr the link $L$ as:
  $$
  R_L^{II}=R_L(p_B(t),R_{B}(t),s(t))^TR_L(p_B(t),R_{B}(t),s_0)
  $$
- **chain calibration**
  **command:** `calibrate <parentLinkName> <childLinkName>`
  in this case the calibration is performed for a chain of links $\mathcal{L}$ from the parent link to the child link. It assumes that all the joint $\mathcal{J}$ involved in the chain are in the zero configuration.
  $$
  \begin{equation}
  \tilde{s}_j=s_j(t) \ \forall j \notin \mathcal{J} \\\\
  \tilde{s}_j=0 \ \ otherwise
  \end{equation}
  $$
  and the secondary calibration is computed for the links in $\mathcal{L}$ as:
  $$
  R_i^{II}=R_i(p_B(t),R_{B}(t),s(t))^TR_i(p_B(t),R_{B}(t),\tilde{s}) \\ \forall i \in \mathcal{L}
  $$

lrapetti:
[...]

PR #274, opened 2021-11-15
Add different target and sensors type in HumanStateProvider

lrapetti:
This PR implement the possibility of using different type of wearable sensors and associate them with different sensors type (https://github.com/robotology/human-dynamics-estimation/issues/273).

Currently the following sensor/targets type have been implemented:
- wearable sensors (see https://github.com/robotology/human-dynamics-estimation/pull/274/commits/9ee3ff1475946c1a446d19209b50ad3c6aec41a0#diff-3daa57ee2a77e48985ef2075ff08d8b318f9e9c67fd813230473f4492cf19fd1R1571-R1650)
  - `PoseSensor`
  - `VirtualLinkKinSensor`
- target types (https://github.com/robotology/human-dynamics-estimation/pull/274/commits/8848414d6498dc8147235ae31c91f956078715f6#diff-3daa57ee2a77e48985ef2075ff08d8b318f9e9c67fd813230473f4492cf19fd1R119-R127)
  - pose,
  - poseAndVelocity,
  - position,
  - positionAndVelocity,
  - orientation,
  - orientationAndVelocity

Before merging:
- [x] Fix and test the calibration methods
- [ ] Fix global IK
- [ ] Fix Pairwised IK
- [ ] Final clean-up and remove deprecated code

After merging:
- Align configuration files in other repos
  - [`component_ifeel`](https://github.com/ami-iit/component_ifeel)


conf/xml/hands-pHRI.xml
```diff
@@ -51,30 +51,32 @@
...
-            <param name="map_RightLowerLeg">(RightLowerLeg, XsensSuit::vLink::RightLowerLeg)</param>
-            <param name="map_RightFoot">(RightFoot, XsensSuit::vLink::RightFoot)</param>
-            <param name="map_RightToe">(RightToe, XsensSuit::vLink::RightToe)</param>
-            <param name="map_LeftUpperLeg">(LeftUpperLeg, XsensSuit::vLink::LeftUpperLeg)</param>
-            <param name="map_LeftLowerLeg">(LeftLowerLeg, XsensSuit::vLink::LeftLowerLeg)</param>
-            <param name="map_LeftFoot">(LeftFoot, XsensSuit::vLink::LeftFoot)</param>
-            <param name="map_LeftToe">(LeftToe, XsensSuit::vLink::LeftToe)</param>
+        <group name="WEARABLE_SENSOR_TARGETS">
+            <!-- LinkName, WearableSensorName, TargetType-->
+            <!-- TargetType : {pose, poseAndVelocity, position, positionAndVelocity, orientation, orientationAndVelocity}-->
```
S-Dafarra, line 56: Picky comment. This line is present only in a couple of these conf files

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -122,17 +117,93 @@ enum rpcCommand
...
+static std::unordered_map<std::string,TargetType> const stringToTargetType = { {"pose",TargetType::pose},
+                                                                               {"poseAndVelocity",TargetType::poseAndVelocity},
+                                                                               {"position",TargetType::position},
+                                                                               {"positionAndVelocity",TargetType::positionAndVelocity},
+                                                                               {"orientation",TargetType::orientation},
+                                                                               {"orientationAndVelocity",TargetType::orientationAndVelocity}};
+
+class WearableSensorTarget
+{
+    public:
```
S-Dafarra, line 139: Picky comment. Usually ``public`` is not indented

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -551,11 +626,27 @@ bool HumanStateProvider::open(yarp::os::Searchable& config)
...
+        auto it = stringToTargetType.find(targetTypeString);
+        if (it == stringToTargetType.end()) {
+            yError() << LogPrefix << "Target Type not valid [" << targetTypeString << "]";
+            return false;
+        }
+        TargetType targetType = it->second;
+        TargetName targetName = linksGroup.get(i).asList()->get(0).asString();
+
+        // yInfo() << LogPrefix << "Read link map:" << modelLinkName << "==>" << wearableName;
+        // pImpl->wearableStorage.modelToWearable_LinkName[modelLinkName] = wearableName;
```
S-Dafarra, line 641: Leftover?

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -1133,44 +1224,51 @@ bool HumanStateProvider::close()
...
+    //     askToStop();
+    //     return;
+    // }
+
+    // // Get the link velocity from input data
+    // if (!pImpl->getLinkVelocityFromInputData(pImpl->linkVelocities)) {
+    //     yError() << LogPrefix << "Failed to get link velocity from input data";
+    //     askToStop();
+    //     return;
+    // }
```
S-Dafarra, line 1271: Leftover?

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -1357,36 +1455,56 @@ void HumanStateProvider::impl::computeSecondaryCalibrationRotationsForChain(cons
...
+
+    // // computing the world calibration
+    // secondaryCalibrationWorld = iDynTree::Transform::Identity();
+    // if (refLinkForCalibrationName!="")
+    // {
+    //     iDynTree::Transform linkForCalibrationTransform = kinDynComputations->getWorldTransform(refLinkForCalibrationName);
+    //     secondaryCalibrationWorld = refLinkForCalibrationTransform * linkForCalibrationTransform.inverse();
+    //     yInfo() << LogPrefix << "secondary calibration for the World is set to " << secondaryCalibrationWorld.toString();
+    // }
```
S-Dafarra, line 1507: Leftover?

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -1423,6 +1541,8 @@ bool HumanStateProvider::impl::applyRpcCommand()
         jointZeroIndices.resize(kinDynComputations->getNrOfDegreesOfFreedom());
         std::iota(jointZeroIndices.begin(), jointZeroIndices.end(), 0);
 
+        std::cerr << "Applying calibrate ALL!" << std::endl;
```
S-Dafarra, line 1544: Probably it is better to use yarp logging here.

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -1357,36 +1455,56 @@ void HumanStateProvider::impl::computeSecondaryCalibrationRotationsForChain(cons
...
+    
+    for (auto wearableTargetEntry : wearableTargets)
+    {
+        TargetName targetName = wearableTargetEntry.first;
+        ModelLinkName linkName = wearableTargetEntry.second->modelLinkName;
+        iDynTree::LinkIndex linkIndex = kinDynComputations->model().getLinkIndex(linkName);
 
-    kinDynComputations->setRobotState(linkTransformMatricesRaw.at(kinDynComputations->getFloatingBase()), jointPos, baseVel, jointVel, worldGravity);
+        std::cerr << "target: " << targetName << std::endl;
+        std::cerr << "link index" << linkIndex << std::endl;
```
S-Dafarra, line 1468: Is this a debug print? If it is meant to stay, probably it would be better using ``yarp`` logging.

devices/HumanStateProvider/HumanStateProvider.cpp
```diff
@@ -122,17 +117,93 @@ enum rpcCommand
...
+};
+
+static std::unordered_map<std::string,TargetType> const stringToTargetType = { {"pose",TargetType::pose},
+                                                                               {"poseAndVelocity",TargetType::poseAndVelocity},
+                                                                               {"position",TargetType::position},
+                                                                               {"positionAndVelocity",TargetType::positionAndVelocity},
+                                                                               {"orientation",TargetType::orientation},
+                                                                               {"orientationAndVelocity",TargetType::orientationAndVelocity}};
+
+class WearableSensorTarget
```
S-Dafarra, line 137: This file is pretty huge. Consider moving the additional classes to separate files

lrapetti:
With https://github.com/robotology/human-dynamics-estimation/pull/274/commits/2d9da1112a027ae522935780bb74f4012054784e I am adding to this PR the possibility to pass from configuration file both a left and a right rotation transform to apply to the measurements. Those can be defined using the following group parameters name:
- `MEASUREMENT_TO_LINK_ROTATIONS` is the right rotation (that is replacing the already existing option `FIXED_SENSOR_ROTATION `)
- `WORLD_TO_MEASUREMENT_ROTATIONS` is the left rotation

lrapetti:
> - [x] Fix and test the calibration methods

This point have been addressed. Note that, however, in https://github.com/robotology/human-dynamics-estimation/pull/274/commits/c1b03635943422e5f23def005a00b11fe7262fa3 I have removed two calibration methods that currently were not used:
- `calibrate <linkName>` (this method is basically equivalent to `calibrateRelativeLink`, the difference is that you have to specify the parent link rather then having it automatically chosen from end-effector pairs)
- `calibrateSubTree` (note also that this method is still not properly working for all the models, see https://github.com/robotology/human-dynamics-estimation/issues/208#issuecomment-666505443)

lrapetti:
> - [ ] Fix Pairwised IK

Since we are now enabling the possibility of having multiple targets for the same links and we are adding also position sensors (that can not be taken into account by using reduced models only), I think we can think of deprecating the pairwise IK methods since it is not straightforward to align it with the new changes.

What do you think @S-Dafarra @kouroshD?

A similar problem exists for the Global IK, we can not define multiple targets for the same link, but in this case I would try to maintain it for the case in which multiple targets are not defined for the same link.

S-Dafarra:
> > * [ ]  Fix Pairwised IK
> 
> Since we are now enabling the possibility of having multiple targets for the same links and we are adding also position sensors (that can not be taken into account by using reduced models only), I think we can think of deprecating the pairwise IK methods since it is not straightforward to align it with the new changes.
> 
> What do you think @S-Dafarra @kouroshD?
> 
> A similar problem exists for the Global IK, we can not define multiple targets for the same link, but in this case I would try to maintain it for the case in which multiple targets are not defined for the same link.

I think that's ok to deprecate it, also because until now we did not find a case where it was working better

interfaces/IWearableTargets/hde/interfaces/IWearableTargets.h
```diff
@@ -0,0 +1,136 @@
...
+            linearVelocityScaled.setVal(0, linearVelocityInWorld.getVal(0) * positionScaleFactor(0));
+            linearVelocityScaled.setVal(1, linearVelocityInWorld.getVal(1) * positionScaleFactor(1));
+            linearVelocityScaled.setVal(2, linearVelocityInWorld.getVal(2) * positionScaleFactor(2));
+            return linearVelocityScaled;
+        };
+
+        iDynTree::Vector3 getCalibratedAngularVelocity()
+        {
+            std::lock_guard<std::mutex> lock(mutex);
+            return iDynTree::AngularMotionVector3(angularVelocity).changeCoordFrame(calibrationWorldToMeasurementWorld.getRotation());
```
lrapetti, line 117: Here we should take into consideration also the offset applied in the right calibration transform

lrapetti:
Closing this PR for the time being. To be reopened once all the required changes are completed.

Issue #277, opened 2022-01-04
Add the possibility to use vector targets in the IK


lrapetti:
In particular, we are interested in aligning the measured gravity vector.

A possible reference for this can be eq.11 from https://www.sciencedirect.com/science/article/pii/S0005109814006104 by @DanielePucci. In a nutshell, the following control law can be used:
$$
\omega = -k (z^{meas} \times z) + \omega^{meas}
$$
with $k>0$

lrapetti:
An initial implementation can be found in https://github.com/robotology/human-dynamics-estimation/commit/21bbb5dac4e887f37502740576f2197245a6faca. The implementation has been done in the `DynamicalInverseKinematics` class but it is not properly used in the `HumanStateProvider` (in that branch the z components only is used for all the orientation targets, instead we should define a new type of target).

lrapetti:
Feature merged onto the main development branch `add-different-target-types`, we can close this issue.

PR #281, opened 2022-01-10
Add gravity vector target in HumanStateProvider

lrapetti:
This PR adds what was discussed in https://github.com/robotology/human-dynamics-estimation/issues/277, by:
- implementing the interfaces for selecting a single component of the rotation in the `DynamicalInverseKinematics` (plus the possibility of selecting also the position components
- on the HumanStateProvider I have created a new target type called `gravity` which select only the z component of the rotation as in the formula in https://github.com/robotology/human-dynamics-estimation/issues/277#issue-1093247974
- in the visualizer, I have add the possibility to visualize such targets by visualizing the gravity vector in the current frame (i.e. $Rz^{meas}$


lrapetti:
This PR is done onto development branch `add-different-target-types`, so I will proceed merging, and then from that branch we will open the PR towards master.

PR #289, opened 2022-03-10
Add parameter for setting all the position scaling factors at once

lrapetti:
This PR adds the possibility to use a single parameter (`x_scale_factor_all`,`y_scale_factor_all`,`z_scale_factor_all `) for the position scale factors for all the targets.

In this way, until https://github.com/robotology/yarp/issues/2812 is not fixed, the scale factors can be passed via command line (as long as we want to set them to the same value for all the targets). 
I have also added the `extern-name` option in the [RobotStateProvider_iCub2_5_openxr_ifeel.xml](https://github.com/robotology/human-dynamics-estimation/compare/xprize...add-scaleFactorAll-flag?expand=1#diff-dc42fd4f001c52fc080e027c8397b453e5817fa76d96e7561ecf82bd440245c1) configuration so that we can set then as:
```sh
yarprobotinterface --config RobotStateProvider_iCub2_5_openxr_ifeel.xml --xy_scale $SCALE_FOR_XY --z_scale $SCALE_FOR_Z
```

### DIFF
--- a/conf/xml/RobotStateProvider_iCub2_5_openxr_ifeel.xml
+++ b/conf/xml/RobotStateProvider_iCub2_5_openxr_ifeel.xml
@@ -101,6 +101,9 @@
             <param name="target_Head">( 0.7 0.7 0.6 )</param>
             <param name="target_LeftHand">( 0.7 0.7 0.6 )</param>
             <param name="target_RightHand">( 0.7 0.7 0.6 )</param>
+            <param name="x_scale_factor_all" extern-name="xy_scale">0.7</param>
+            <param name="y_scale_factor_all" extern-name="xy_scale">0.7</param>
+            <param name="z_scale_factor_all" extern-name="z_scale">0.6</param>
         </group>
         <group name="CUSTOM_CONSTRAINTS">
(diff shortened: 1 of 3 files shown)


lrapetti:
@S-Dafarra after merging this, we can test it with the full system working and add it to the `yarpmanager` configuration

lrapetti:
> @S-Dafarra after merging this, we can test it with the full system working and add it to the `yarpmanager` configuration

I have already opened the PR for the `yarpmanager` application at https://github.com/robotology/walking-teleoperation/pull/93/files

S-Dafarra (approved):
I did not spot anything weird (apart from the spacing in https://github.com/robotology/human-dynamics-estimation/blob/e97fa1b00b09769331ccfe27377cc87532bad80d/devices/HumanStateProvider/HumanStateProvider.cpp#L570-L573)

Also I am not expert of the code, but overall it looks good to me!

lrapetti:
Concerning the spacing, it seems the `clang-format` didn't work out. Anyway I'll proceed merging since we might not even want to port this change as it is in main development branch in the future

PR #323, opened 2022-10-21
Make dependency on robometry optional and add it in the documentation

lrapetti:
In https://github.com/robotology/human-dynamics-estimation/pull/298 we forgot to update the documentation with the `robometry` dependency.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -212,6 +212,7 @@ jobs:
             cmake -A x64 cmake -G"Visual Studio 17 2022" -DCMAKE_PREFIX_PATH=${GITHUB_WORKSPACE}/install/deps \
                                                          -DCMAKE_BUILD_TPYE=${{matrix.build_type}} \
                                                          -DHUMANSTATEPROVIDER_ENABLE_VISUALIZER:BOOL=ON \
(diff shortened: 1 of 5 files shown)


lrapetti:
Note that the library is still marked as **Under Development**. We might think of making this dependency optional

traversaro (approved):

RiccardoGrieco:
> Note that the library is still marked as **Under Development**. We might think of making this dependency optional

If we want to make it optional we should address it also in the device [CMakeLists.txt](https://github.com/robotology/human-dynamics-estimation/blob/3384a0bd6b8411c865e38136a30ece69241ce7c7/devices/HumanLogger/CMakeLists.txt#L6)

lrapetti:
> > Note that the library is still marked as **Under Development**. We might think of making this dependency optional
> 
> If we want to make it optional we should address it also in the device [CMakeLists.txt](https://github.com/robotology/human-dynamics-estimation/blob/3384a0bd6b8411c865e38136a30ece69241ce7c7/devices/HumanLogger/CMakeLists.txt#L6)

Yess, it has been addressed in https://github.com/robotology/human-dynamics-estimation/pull/323/commits/165183da985950e8524b091814c96ba8825ce4cd

lrapetti:
@RiccardoGrieco do you have any further comment or can we merge it?

RiccardoGrieco (approved):

PR #326, opened 2022-11-26
Fix iCub3 retargeting application using openXR

lrapetti:
This PR fix the configuration files for running the retargeting on iCub3 using openxr


lrapetti:
Thanks @S-Dafarra

PR #338, opened 2023-02-15
Fix HumanLogger not working when logging just one quantity

RiccardoGrieco:
The periodic thread (e.g. the `run()` function) in human logger is started inside the `attachAll` method.

Turns out that, since the device implements both `IWrapper` and `IMultipleWrapper`, the first one is preferred when the list of devices to attach contains only one of them (see [here](https://github.com/robotology/yarp/blob/e845b1de554f4bf67c262de9794ebbc2bf9ca29f/src/libYARP_robotinterface/src/yarp/robotinterface/Device.cpp#L416-L451)).

In such case, only the `attach()` method belonging to `IWrapper` is called. 
In our case, this happens when using the logging of only one between the kinematics and dynamics data.

This PR is to solve this issue.

### DIFF
--- a/devices/HumanLogger/HumanLogger.cpp
+++ b/devices/HumanLogger/HumanLogger.cpp
@@ -47,8 +47,6 @@ struct hde::devices::HumanLoggerSettings
 
 using namespace hde::devices;
 
-using MatlabChannelName = std::string;
-
 class HumanLogger::impl
 {
 public:
@@ -61,7 +59,6 @@ public:
                                    bool& option);
     bool configureBufferManager();
 
-    bool firstRun = true;
     size_t waitingFirstReadCounter = 1;
 
     interfaces::IHumanState* iHumanState = nullptr;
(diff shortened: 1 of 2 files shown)


RiccardoGrieco:
cc @claudia-lat

devices/HumanLogger/HumanLogger.cpp
```diff
@@ -353,6 +355,20 @@ bool HumanLogger::attach(yarp::dev::PolyDriver* poly)
...
+    // so we have to start the periodic thread from here
+    if(pImpl->numberOfDevicesToAttach == 1)
+    {
+        if (!pImpl->configureBufferManager()) {
+            yError() << logPrefix << "Failed to configure buffer manager for the logger.";
+        }
+
+        if(!start()) {
+            yError() << logPrefix << "Failed to start the loop.";
+            return false;
```
lrapetti, line 368: This part is duplicated in the two methods, what do you think of creating a method to avoid duplication.

lrapetti (approved):

devices/HumanLogger/HumanLogger.cpp
RiccardoGrieco: You are right. I changed the modification moving all the start logic only into the `attach()` method

lrapetti (approved):
Thanks @RiccardoGrieco, for me we can proceed merging.

PR #341, opened 2023-03-07
Added custom settings for calibration and visualization options

dariosortino:
As per the title, with @S-Dafarra we included the possibility to define window's Height and Width from the configuration file of the Human State Visualizer. Also, with @lrapetti and @mebbaid we define a new parameter called CalibrationConfig that allows setting custom angles for calibration purposes.

PR #348, opened 2023-04-04
Fix calibration using jointCalibrationSolution

lrapetti:
In https://github.com/robotology/human-dynamics-estimation/pull/341, after the code review we have introduced a bug and the calibration was no longer performed correctly.

The problem was that the `jointCalibrationSolution` was never used, while the current configuration (`jointConfigurationSolution`) was used to calibrate the joints. This PR fix the bug


RiccardoGrieco (approved):
LGTM

Issue #356, opened 2023-09-28
[HumanLogger] logger not starting because interface is not ready


lrapetti:
The `HumanLogger` sometimes do not start because the interface is not ready.

```
...
[DEBUG] |yarp.dev.PolyDriver|HumanLoggerDevice| Parameters are (LoggerType matlab) (auto_save true) (data_threshold 300) (device human_logger) (experimentName human_data) (id HumanLoggerDevice) (logHumanDynamics true) (logHumanState true) (n_samples 100000) (path "experiment_dir/") (period 0.0100000000000000002082) (robotName) (saveBufferManagerConfiguration false) (save_period 120.0) (save_periodically true)
[INFO] HumanLogger : LoggerType set to MATLAB
[INFO] |yarp.dev.PolyDriver|HumanLoggerDevice| Created device <human_logger>. See C++ class hde::devices::HumanLogger for documentation.
[INFO] Entering action level 5 of phase startup
[INFO] Executing attach action, level 5 on device HumanLoggerDevice with parameters [("networks" = "(HumanStateRemapper HumanDynamicsRemapper)"), ("HumanStateRemapper" = "HumanStateRemapper"), ("HumanDynamicsRemapper" = "HumanDynamicsRemapper")]
[INFO] HumanLogger : Device human_state_remapper successfully attached
[ERROR] HumanLogger : The IHumanDynamics interface human_dynamics_remapper might not be ready
[ERROR] Device HumanLoggerDevice cannot execute attach
[ERROR] Cannot run attach action on device HumanLoggerDevice
[INFO] All actions for action level 5 of startup phase started. Waiting for unfinished actions.
[INFO] All actions for action level 5 of startup phase finished.
[WARNING] There was some problem running actions for startup phase . Please check the log and your configuration
[INFO] startup phase finished.
[ERROR] Error in startup phase... see previous messages for more info
[WARNING] Interrupt # 1 # received.
[INFO] Interrupt received. Stopping all running threads.
[INFO] interrupt1 phase starting...
[INFO] interrupt1 phase finished.
[INFO] shutdown phase starting...
[INFO] Entering action level 5 of phase shutdown
[INFO] Executing detach action, level 5 on device HumanLoggerDevice with parameters []
[INFO] All actions for action level 5 of shutdown phase started. Waiting for unfinished actions.
[INFO] All actions for action level 5 of shutdown phase finished.
[INFO] Closing device HumanLoggerDevice
[...]

PR #357, opened 2023-09-28
Add timeout to wait for data available in interface

lrapetti:
With this PR I am adding a timeout to wait for data being available on the interface. This avoid problems such as https://github.com/robotology/human-dynamics-estimation/issues/356.

Moreover, in https://github.com/robotology/human-dynamics-estimation/commit/7ae0be3fd337b7042c4d82e35640080c90665334 I standardizing the capitalization for the `LogPrefix` variable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - The license of the repo changed to [`BSD-3-Clause`](https://spdx.org/licenses/BSD-3-Clause.html) (https://github.com/robotology/human-dynamics-estimation/pull/350).
+- Devices wait for data available in IHumanState and IHumanDynamics interfaces instead of returning error (https://github.com/robotology/human-dynamics-estimation/pull/357).
 
 ### Fixed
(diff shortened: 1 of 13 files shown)


lrapetti:
@traversaro @GiulioRomualdi @davidegorbani I don't know if this solution make sense to you, or you can propose alternatives? (clearly we might add a parameter that defines the timeout to avoid the device getting stuck waiting)

traversaro (approved):
As a general comment, it probably make sense to extend the IHumanState interface to also provide this information about the status, as conflating getNumberOfJoints == 0 ==> no first data received may be a bit problematic. Anyhow, as a first fix this is perfect.

lrapetti:
> As a general comment, it probably make sense to extend the IHumanState interface to also provide this information about the status, as conflating getNumberOfJoints == 0 ==> no first data received may be a bit problematic. Anyhow, as a first fix this is perfect.

That make sense, I can open an issue related to this. As usual, a concern is that it would imply that we should update the associated thrift message, and old datasets might no longer be compatible (however, I don't think we are directly using `IHumanState` dumped signal so it should not be a problem).

lrapetti:
> > As a general comment, it probably make sense to extend the IHumanState interface to also provide this information about the status, as conflating getNumberOfJoints == 0 ==> no first data received may be a bit problematic. Anyhow, as a first fix this is perfect.
> 
> That make sense, I can open an issue related to this.

Done https://github.com/robotology/human-dynamics-estimation/issues/360, I'll proceed updating the `CHANGELOG.md` and merging.

PR #383, opened 2024-03-05
Update of configuration files

S-Dafarra:


traversaro (approved):

mebbaid:
Hi @S-Dafarra , thanks !
Do you think we should update this file as well https://github.com/robotology/human-dynamics-estimation/blob/master/conf/xml/transformServer_to_iwear.xml ?

mebbaid (approved):

mebbaid:
>  Do you think we should update this file as well https://github.com/robotology/human-dynamics-estimation/blob/master/conf/xml/transformServer_to_iwear.xml ?

On a second thought, since I am only using it with data dumpers, we can proceed with merge. Thanks

PR #421, opened 2024-12-04
Set the calibrationMeasurementToLink matrix to the initial when resetAll is called

davidegorbani:
As per title, with this PR the calibration matrix that $^{link}R_{measurement}$ is set to the initial value read from the configuration file when the rpc command `resetAll` is called.

CC @LudovicaDanovaro @dariosortino


dariosortino (approved):
Amazing, thanks

traversaro (approved):

PR #433, opened 2026-06-04
Add autoreconnect to network clients

RiccardoGrieco:
This pull request adds automatic reconnection logic for their input ports if the connection to the expected data source is lost.

This allows e.g. the HumanStateVisualizer to keep running while the other modules restart 

Key changes include:

**Automatic reconnection and port monitoring:**
* Each client (`HumanDynamics_nwc_yarp`, `HumanState_nwc_yarp`, `HumanWrench_nwc_yarp`, `WearableTargets_nwc_yarp`) now checks in its `run()` method whether its input port is disconnected. If so, it logs a warning (throttled to once every 5 seconds) and attempts to reconnect to the configured data port. This makes the clients more resilient to temporary network issues.

cc @camillagallina2000


S-Dafarra:
Consider that even if the network drops, when it is available again, the communication should restart without issues, provided that the sockets did not close.

This is useful in case the server is restarted. Nonetheless, I would avoid doing this connections automatically. I would suggest to use a RPC instead. The reasons are:
- in the past we noticed a race condition on ``getOutputCount``: https://github.com/robotology/yarp/issues/2531. It is possible this very same bug is still there and it can also affect ``getInputCount`` considering that is its dual. The race condition can occur if you are checking the number of inputs in the very same moment one of those inputs is disappearing
- direct requests to the name server. Especially if you are in a degraded network, asking continuously the name server to do a connection makes things worse. 

Hence, if the reconnection is done "manually", you can avoid both issues. 

Another possibility would be to enable this mechanism only via optional configuration parameter

traversaro (approved):
Ok for me once @S-Dafarra is happy.

RiccardoGrieco:
@S-Dafarra I switched to using yarp's `PortReport` to check for disconnections. Also I made the reconnection enabled by a configuration parameter

S-Dafarra (approved):
Looks good! Thanks

PR #436, opened 2026-09-08
Add HUMANSTATEPROVIDER_ENABLE_YARP option to disable part of hde that depend on YARP

traversaro:
The `IWear` and hde interfaces are already YARP-independent, so they can be use directly with dinrail, so I think it make sense to have an option to install only the interfaces without everything else.


CMakeLists.txt
RiccardoGrieco: This requires YARP to build, so I would move it inside the `if(HUMANSTATEPROVIDER_ENABLE_YARP)` block to avoid unnecessary find_package calls.

RiccardoGrieco: All these require YARP to build, so I would move them inside `if(HUMANSTATEPROVIDER_ENABLE_YARP)`  block

RiccardoGrieco (changes requested):

CMakeLists.txt
traversaro: Done.

traversaro: Done.

RiccardoGrieco (approved):

